PR #23, opened 2021-07-06
Allow vmi pause state change timeout to include initial CRD state

sarahbx:
##### Short description:
This change allows the state change from paused-to-unpaused or vice-versa be with in the timeout period.

### DIFF
--- a/ocp_resources/virtual_machine.py
+++ b/ocp_resources/virtual_machine.py
@@ -292,6 +292,15 @@ class VirtualMachineInstance(NamespacedResource):
             func=self.get_vmi_active_condition,
         )
         for sample in samples:
+            # VM in state change
+            # We have commanded a [un]pause condition via the API but the CR has not been updated yet to match.
+            # 'reason' may not exist yet
+            # or
+            # 'reason' may still exist after unpause if the CR has not been updated before we perform this check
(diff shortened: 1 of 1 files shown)


ocp_resources/virtual_machine.py
```diff
@@ -292,6 +292,11 @@ def wait_for_vmi_condition_pause_status(self, pause, timeout=TIMEOUT):
             func=self.get_vmi_active_condition,
         )
         for sample in samples:
+            # VM in state change
+            if (pause and not sample.get("reason")) or (
+                sample.get("reason") == "PausedByUser" and not pause
+            ):
+                continue
```
myakove, line 299: Sorry, but I don't understand what should be the results here
Can you please add comments for each and explain when and why this is needed?

maybe somthing like:
xxx = pause and not sample.get("reason")  # explain ......
yyy = sample.get("reason") == "PausedByUser"  # explain ......

and please elaborate more about "# VM in state change"

myakove (changes requested):

ocp_resources/virtual_machine.py
sarahbx: Comment updated

myakove (approved):

myakove (approved):

PR #24, opened 2021-07-06
TimeoutSampler exceptions arg deprecation for Resource get() retry

sarahbx:
```release-note
    TimeoutSampler exceptions arg deprecation for Resource get() retry
    
    TimeoutSampler:
    - Deprecated: 'exceptions' and 'exceptions_msg' arguments will be removed
    - New: exceptions_dict argument to handle additional exception states
    
    Resource:
    - Changed: additional exception states retried in get() to increase resiliency
```

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -30,4 +30,4 @@ jobs:
         python -m pip install tox
     - name: Run tox
       run: |
-       tox -e code-check
+       tox -e code-check,unittests
--- a/ocp_resources/utils.py
+++ b/ocp_resources/utils.py
@@ -1,8 +1,13 @@
 import logging
 import time
+from warnings import warn
 
 
 LOGGER = logging.getLogger(__name__)
+
+
+class InvalidArgumentsError(Exception):
+    pass
 
 
 class TimeoutExpiredError(Exception):
@@ -25,6 +30,47 @@ class TimeoutSampler:
     Yielding the output allows you to handle every value as you wish.
 
     Feel free to set the instance variables.
+
+    exceptions_dict should be in the following format:
+    {
+        exception0: [exception0_msg0],
+        exception1: [
+            exception1_msg0,
+            exception1_msg1
+        ],
+        exception2: []
+    }
+
+    If an exception is raised within `func`:
+        Example exception inheritance:
+            class Exception
+            class AExampleError(Exception)
+            class BExampleError(AExampleError)
+
+        The raised exception's class will fall into one of three categories:
+            1. An exception class specifically declared in exceptions_dict
+                exceptions_dict: {BExampleError: []}
+                raise: BExampleError
+                result: continue
+
+            2. A child class inherited from an exception class in exceptions_dict
+                exceptions_dict: {AExampleError: []}
+                raise: BExampleError
+                result: continue
+
+            3. Everything else, this will always re-raise the exception
+                exceptions_dict: {BExampleError: []}
+                raise: AExampleError
+                result: raise
+
+    Args:
+        wait_timeout (int): Time in seconds to wait for func to return a value equating to True
+        sleep (int): Time in seconds between calls to func
+        func (function): to be wrapped by TimeoutSampler
+        exceptions (tuple): Deprecated: Tuple containing all retry exceptions
+        exceptions_msg (str): Deprecated: String to match exception against
+        exceptions_dict (dict): Exception handling definition
+        print_log (bool): Print elapsed time to log
     """
 
     def __init__(
@@ -32,8 +78,9 @@ class TimeoutSampler:
         wait_timeout,
         sleep,
         func,
-        exceptions=None,
-        exceptions_msg=None,
+        exceptions=None,  # TODO: Deprecated and will be removed, use exceptions_dict
+        exceptions_msg=None,  # TODO: Deprecated and will be removed, use exceptions_dict
+        exceptions_dict=None,
         print_log=True,
         **func_kwargs,
     ):
@@ -41,60 +88,162 @@ class TimeoutSampler:
         self.sleep = sleep
         self.func = func
         self.func_kwargs = func_kwargs
-        self.exception = exceptions or Exception
         self.elapsed_time = None
-        self.exceptions_msg = exceptions_msg
         self.print_log = print_log
+
+        # TODO: when exceptions arg removed replace with: self.exceptions_dict = exceptions_dict or {Exception: []}
+        self.exceptions_dict = self._pre_process_exceptions(
+            exceptions=exceptions,
+            exceptions_msg=exceptions_msg,
+            exceptions_dict=exceptions_dict,
+        )
+        self._exceptions = tuple(self.exceptions_dict.keys())
+
+    def _pre_process_exceptions(self, exceptions, exceptions_msg, exceptions_dict):
+        """
+        Convert any deprecated `exceptions` and `exceptions_msg` arguments to an `exceptions_dict`
+
+        TODO: Deprecation: This method should be removed when the 'exceptions' argument is removed from __init__
+
+        Args:
+            exceptions (tuple): Deprecated: Tuple containing all retry exceptions
+            exceptions_msg (str): Deprecated: String to match exception against
+            exceptions_dict (dict): Exception handling definition
+
+        Returns:
+            dict: exceptions_dict compatible input
+        """
+        output = {}
+        if exceptions_dict and (exceptions or exceptions_msg):
+            raise InvalidArgumentsError(
+                "Must specify either exceptions_dict or exceptions/exception_msg, not both"
+            )
+
+        elif exceptions or exceptions_msg:
+            warn(
+                "TimeoutSampler() exception and exception_msg are now deprecated. "
+                "Please update to use exceptions_dict by Oct 12, 2021",
+                DeprecationWarning,
+            )
+
+        if exceptions_dict:
+            output.update(exceptions_dict)
+
+        elif exceptions is None:
+            output[Exception] = []
+
+        else:
+            if not isinstance(exceptions, tuple):
+                exceptions = (exceptions,)
+
+            for exp in exceptions:
+                if exceptions_msg:
+                    output[exp] = [exceptions_msg]
+                else:
+                    output[exp] = []
+
+        return output
 
     @property
     def _func_log(self):
+        """
+        Returns:
+            string: `func` information to include in log message
+        """
         return f"Function: {self.func} Kwargs: {self.func_kwargs}"
 
     def __iter__(self):
-        last_exp = None
+        """
+        Iterator
+
+        Yields:
+            any: Return value from `func`
+        """
         timeout_watch = TimeoutWatch(timeout=self.wait_timeout)
         if self.print_log:
             LOGGER.info(
                 f"Waiting for {self.wait_timeout} seconds, retry every {self.sleep} seconds"
             )
 
+        last_exp = None
         while timeout_watch.remaining_time() > 0:
             try:
                 self.elapsed_time = self.wait_timeout - timeout_watch.remaining_time()
                 yield self.func(**self.func_kwargs)
                 self.elapsed_time = None
+
                 time.sleep(self.sleep)
 
-            except self.exception as exp:
+            except self._exceptions as exp:
                 last_exp = exp
-                self.elapsed_time = None
+                last_exp_log = self._get_exception_log(exp=last_exp)
+                if self._is_raisable_exception(exp=last_exp):
+                    LOGGER.error(last_exp_log)
+                    raise exp
 
+                self.elapsed_time = None
                 time.sleep(self.sleep)
 
             finally:
                 if self.elapsed_time and self.print_log:
                     LOGGER.info(f"Elapsed time: {self.elapsed_time}")
 
-        raise TimeoutExpiredError(self._process_execution(exp=last_exp))
+        raise TimeoutExpiredError(self._get_exception_log(exp=last_exp))
 
-    def _process_execution(self, exp=None):
+    @staticmethod
+    def _is_exception_matched(exp, exception_messages):
+        """
+        Args:
+            exp (Exception): Exception object raised by `func`
+            exception_messages (list): Either an empty list allowing all text,
+                                       or a list of allowed strings to match against the exception text.
+
+        Returns:
+            bool: True if exception text is allowed or no exception text given, False otherwise
+        """
+        if not exception_messages:
+            return True
+
+        for msg in exception_messages:
+            # Prevent match if provided with empty string
+            if msg and msg in str(exp):
+                return True
+
+        return False
+
+    def _is_raisable_exception(self, exp):
+        """
+        Verify whether exception should be raised during execution of `func`
+
+        Args:
+            exp (Exception): Exception object raised by `func`
+
+        Returns:
+            bool: True if exp should be raised, False otherwise
+        """
+
+        for entry in self.exceptions_dict:
+            if isinstance(exp, entry):  # Check inheritance for raised exception
+                exception_messages = self.exceptions_dict.get(entry)
+                if self._is_exception_matched(
+                    exp=exp, exception_messages=exception_messages
+                ):
+                    return False
+
+        return True
+
+    def _get_exception_log(self, exp):
+        """
+        Args:
+            exp (any): Raised exception
+
+        Returns:
+            string: Log message for exception
+        """
         exp_name = exp.__class__.__name__ if exp else "N/A"
+
         last_exception_log = f"Last exception: {exp_name}: {exp}"
-
-        log = "{timeout}\n{func_log}\n{last_exception_log}".format(
-            timeout=self.wait_timeout,
-            func_log=self._func_log,
-            last_exception_log=last_exception_log,
-        )
-
-        if self.exceptions_msg:
-            if self.exceptions_msg not in str(exp):
-                LOGGER.error(log)
-                raise exp
-            else:
-                LOGGER.warning(f"{self.exceptions_msg}: Retrying")
-
-        return log
+        return f"{self.wait_timeout}\n{self._func_log}\n{last_exception_log}"
 
 
 class TimeoutWatch:
--- a/tests/unittests/test_utils.py
+++ b/tests/unittests/test_utils.py
@@ -1,0 +1,303 @@
+import re
+import warnings
+
+import pytest
+
+from ocp_resources.utils import (
+    InvalidArgumentsError,
+    TimeoutExpiredError,
+    TimeoutSampler,
+)
+
+
+class TestTimeoutSampler:
+    @staticmethod
+    def _raise_exception(runtime_exception):
+        if runtime_exception:
+            raise runtime_exception
+
+    def _trigger_func_exception_during_iter(self, exceptions_dict, runtime_exception):
+        for _ in TimeoutSampler(
+            wait_timeout=1,
+            sleep=1,
+            func=self._raise_exception,
+            exceptions_dict=exceptions_dict,
+            print_log=False,
+            runtime_exception=runtime_exception,
+        ):
+            continue
+
+    @pytest.mark.parametrize(
+        "test_params, expected",
+        [
+            pytest.param(
+                {
+                    "init_exceptions_dict": {
+                        KeyError: [],
+                    },
+                    "runtime_exception": ValueError(),
+                },
+                {
+                    "raises": ValueError,
+                },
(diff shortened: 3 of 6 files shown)


ocp_resources/utils.py
```diff
@@ -26,30 +27,97 @@ class TimeoutSampler:
...
+        exceptions_msg=None,  # Deprecated
+        exceptions_dict=None,
         print_log=True,
+        *func_args,
         **func_kwargs,
     ):
         self.wait_timeout = wait_timeout
         self.sleep = sleep
         self.func = func
+        self.func_args = func_args
```
rnetser, line 66: Please rebase, func_args was removed (was not used and func_kwargs should be used if needed)

sarahbx: corrected

ocp_resources/utils.py
```diff
@@ -26,27 +27,90 @@ class TimeoutSampler:
...
+            self.exceptions_dict.update(exceptions_dict)
+            self.exception = tuple(self.exceptions_dict.keys())
+
+        elif exceptions is None:
+            self.exceptions_dict[Exception] = []
+            self.exception = (Exception,)
+
+        else:
+            self.exception = exceptions
+            self.exceptions_msg = exceptions_msg
```
myakove, line 112: You need to process all the args (exception, exception_msg) and create only a dict and work with it
in this function you should return only exceptions_dict

myakove (changes requested):

ocp_resources/utils.py
rnetser: @sarahbx as discussed, this will require a refactor. Let's refactor the existing code to use only the dict (and remove self.exception usage)

sarahbx: @myakove @rnetser as mentioned previously, we are retaining `self.exception` so we only process the tuple of extensions once. I was not intending to reverse process the input arguments exceptions/exception_msg into exceptions_dict as this code will be removed when exceptions/exceptions_msg arguments are removed.

rnetser: Sure but it is not needed as a separate variable; it is only used once and thus:

`except tuple(self.exceptions_dict.keys()) as exp:`

sarahbx: `except tuple(self.exceptions_dict.keys()) as exp:` is nested within a `while`, nested within an `iter`. It will be run more than once.

sarahbx: Code updated. Removed `self.exceptions_msg` and only use `self.exceptions_dict` and `self._exceptions` (to ensure we only calculate it once per object)

ocp_resources/utils.py
```diff
@@ -26,27 +27,93 @@ class TimeoutSampler:
...
         self.wait_timeout = wait_timeout
         self.sleep = sleep
         self.func = func
         self.func_kwargs = func_kwargs
-        self.exception = exceptions or Exception
         self.elapsed_time = None
-        self.exceptions_msg = exceptions_msg
         self.print_log = print_log
 
+        self.exceptions_dict = {}
```
rnetser, line 69: `self.exceptions_dict = exceptions_dict or {}
`

And please add TODO once deprecated, once _pre_process_exceptions is removed, should be:

self.exceptions_dict = exceptions_dict or {Exception: []}

sarahbx: I will add the todo, but it will remain `self.exceptions_dict = {}` until pre-process is removed.

sarahbx:
Moved to draft temporarily. I am going to add a test for `_process_execution()` before we continue with this change.

sarahbx:
After https://github.com/RedHatQE/openshift-python-wrapper/pull/53 is merged, will update unit test for `_process_execution` to run `_process_exceptions` with additional test cases.

ocp_resources/utils.py
```diff
@@ -26,74 +27,154 @@ class TimeoutSampler:
...
                 self.elapsed_time = None
-
                 time.sleep(self.sleep)
 
             finally:
                 if self.elapsed_time and self.print_log:
                     LOGGER.info(f"Elapsed time: {self.elapsed_time}")
 
-        raise TimeoutExpiredError(self._process_execution(exp=last_exp))
+        raise TimeoutExpiredError(last_exp_log)
```
myakove, line 148:
```suggestion
        raise TimeoutExpiredError(_process_exeception(exp=exp))

```

myakove (changes requested):

tests/test_utils.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.mark.parametrize(
+    "test_params, expected",
+    [
+        pytest.param(
+            {},
+            {
+                "exceptions": (Exception,),
+                "exceptions_dict": {Exception: []},
+            },
```
myakove, line 16: Please add id= to the test, it's easier to identify and debug when needed.

myakove (changes requested):

tests/test_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        ),
+    ],
+)
+def test_timeout_sampler_pre_process_exceptions(test_params, expected):
+    def _timeout_sampler():
+        return TimeoutSampler(
+            wait_timeout=1,
+            sleep=1,
+            func=None,
```
myakove, line 107: Please send the right type here, can be send lambda: True

tests/test_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            exceptions=test_params.get("exceptions"),
+            exceptions_msg=test_params.get("exceptions_msg"),
+            exceptions_dict=test_params.get("exceptions_dict"),
+            print_log=None,
```
myakove, line 111: wither True of False.

tests/test_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+
+    if expected.get("raises"):
+        with pytest.raises(expected["raises"]):
+            _timeout_sampler()
+    else:
+        with warnings.catch_warnings(record=True) as w:
```
myakove, line 118: Please use meaningful names and not one latter word.

myakove (changes requested):

tests/test_utils.py
```diff
@@ -0,0 +1,129 @@
+import warnings
```
myakove, line 1: Please add the tests to tox.ini

myakove (changes requested):

tests/test_utils.py
sarahbx: Is this not handled by?
https://github.com/RedHatQE/openshift-python-wrapper/blob/master/tox.ini#L24

ocp_resources/utils.py
```diff
@@ -26,74 +27,154 @@ class TimeoutSampler:
...
+        LOGGER.error(log)
+        raise _exp
 
+    def _process_exeception(self, exp=None):
+        if exp and hasattr(exp, "__name__"):
+            # Exceptions that inherit from Exception
+            exp_name = exp.__name__
+        elif exp:
+            # Exceptions that inherit from object eg: AttributeError
+            exp_name = exp.__class__.__name__
```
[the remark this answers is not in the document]
sarahbx: why? If we already know our state, there's no reason to raise an exception to continue the logic path

tests/test_utils.py
myakove: No, we run tox -e code-check
https://github.com/RedHatQE/openshift-python-wrapper/blob/master/.github/workflows/code-check.yml#L33

you should create unittests section and add it to the workflow file
tox -e unittests

ocp_resources/utils.py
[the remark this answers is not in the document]
myakove: better for reading, less lines
up to you

ocp_resources/utils.py
```diff
@@ -26,74 +27,154 @@ class TimeoutSampler:
...
+    @staticmethod
+    def _check_exp_msgs(msgs, _exp, log):
+        for msg in msgs:
+            if msg in str(_exp):
+                return
+
+        LOGGER.error(log)
+        raise _exp
 
+    def _process_exeception(self, exp=None):
```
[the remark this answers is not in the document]
sarahbx: Using `None` as the default argument removes the need for additional `if exp != "N/A"` checks in remaining conditionals.

ocp_resources/utils.py
sarahbx: Does this apply in the new configuration?

tests/unittests/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+            wait_timeout=1,
+            sleep=1,
+            func=_raise_exception,
+            exceptions_dict=test_params.get("init_exceptions_dict"),
+            print_log=False,
+            runtime_exception=test_params.get("runtime_exception"),
+        ):
+            continue
+
+    if expected.get("raises"):
```
myakove, line 253: I would split it into 2 tests.
One that handles `raises` and one for `TimeoutExpiredError` exception check.

myakove (changes requested):

ocp_resources/utils.py
```diff
@@ -26,74 +29,170 @@ class TimeoutSampler:
...
+            bool: True if exception should be raised, False otherwise
+        """
+
+        if exp.__class__ in self.exceptions_dict:
+            if self._message_match(exp=exp, exp_class=exp.__class__):
+                return False
+
+        else:
+            for entry in self.exceptions_dict:
+                if isinstance(exp, entry) and self._message_match(
```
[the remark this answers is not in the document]
sarahbx: This is related to the purpose of these tests, which preserve the original behavior:
https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r678381822

ocp_resources/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class OpenshiftPythonWrapperError(Exception):
```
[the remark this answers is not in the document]
sarahbx: np, I will get this class to stick eventually ;)

ocp_resources/utils.py
```diff
@@ -26,74 +29,170 @@ class TimeoutSampler:
...
 
     def __init__(
         self,
         wait_timeout,
         sleep,
         func,
-        exceptions=None,
-        exceptions_msg=None,
+        exceptions=None,  # Deprecated
+        exceptions_msg=None,  # Deprecated
```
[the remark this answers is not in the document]
sarahbx: I believe this should remain until the agreed deprecation date.

tests/unittests/test_utils.py
```diff
@@ -0,0 +1,269 @@
+import re
+import warnings
+
+import pytest
+
+from ocp_resources.exceptions import InvalidArgumentsError
+from ocp_resources.utils import TimeoutExpiredError, TimeoutSampler
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
sarahbx: I'd prefer we left it until the code it exercises is removed

ocp_resources/utils.py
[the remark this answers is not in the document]
sarahbx: In fact, making exceptions_dict an `OrderedDict` would allow for a consistent exception list order, and enable priorities between different classes/messages.

ocp_resources/utils.py
```diff
@@ -26,74 +29,170 @@ class TimeoutSampler:
...
-            if self.exceptions_msg not in str(exp):
-                LOGGER.error(log)
-                raise exp
-            else:
-                LOGGER.warning(f"{self.exceptions_msg}: Retrying")
+        Returns:
+            bool: True if exception should be raised, False otherwise
+        """
+
+        if exp.__class__ in self.exceptions_dict:
```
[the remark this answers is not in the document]
sarahbx: Related to: https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r678349850

ocp_resources/utils.py
```diff
@@ -26,74 +29,170 @@ class TimeoutSampler:
...
                 if self.elapsed_time and self.print_log:
                     LOGGER.info(f"Elapsed time: {self.elapsed_time}")
 
-        raise TimeoutExpiredError(self._process_execution(exp=last_exp))
+        raise TimeoutExpiredError(last_exp_log)
 
-    def _process_execution(self, exp=None):
-        exp_name = exp.__class__.__name__ if exp else "N/A"
-        last_exception_log = f"Last exception: {exp_name}: {exp}"
+    def _message_match(self, exp, exp_class):
```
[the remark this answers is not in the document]
sarahbx: Related to: https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r678347593

ocp_resources/utils.py
```diff
@@ -26,74 +29,170 @@ class TimeoutSampler:
...
-        if self.exceptions_msg:
-            if self.exceptions_msg not in str(exp):
-                LOGGER.error(log)
-                raise exp
-            else:
-                LOGGER.warning(f"{self.exceptions_msg}: Retrying")
+        Returns:
+            bool: True if exception should be raised, False otherwise
+        """
+
```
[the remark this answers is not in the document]
sarahbx: I will move setting of `exception_messages` to `_is_raisable_exception()`.

However, `_message_match(exp, exception_messages)` needs to be where this logic exists to keep the conditionals in `_is_raisable_exception()` simple and straight-forward.

ocp_resources/utils.py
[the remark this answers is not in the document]
myakove: I don't think we need priorities for  exceptions, all  exceptions have the same priority

ocp_resources/utils.py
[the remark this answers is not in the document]
rnetser: It is no longer used, we should not keep it (i.e like we removed any other un-used code).

ocp_resources/utils.py
```diff
@@ -26,76 +31,178 @@ class TimeoutSampler:
...
+        ],
+        exception2: []
+    }
+
+    Args:
+        wait_timeout (int): Time in seconds to wait for func to return a value equating to True
+        sleep (int): Time in seconds between calls to func
+        func (function): to be wrapped by TimeoutSampler
+        exceptions (tuple): Deprecated: Tuple containing all retry exceptions to pass to TimeoutSampler
+        exceptions_msg (str): Deprecated: String to match exception against
```
rnetser, line 50: Adding my comment here so it's not lost in the thread; `exceptions_msg` should be removed (the same way un-used cod is removed)

ocp_resources/utils.py
```diff
@@ -26,76 +31,178 @@ class TimeoutSampler:
...
 
-        raise TimeoutExpiredError(self._process_execution(exp=last_exp))
+        raise TimeoutExpiredError(last_exp_log)
 
-    def _process_execution(self, exp=None):
-        exp_name = exp.__class__.__name__ if exp else "N/A"
-        last_exception_log = f"Last exception: {exp_name}: {exp}"
+    @staticmethod
+    def _is_exception_message_matched(exp, exception_messages):
+        if not exception_messages:
```
rnetser, line 162: `_is_exception_message_matched` implies that something is looked for.
It should be called only if there're `exception_messages`

ocp_resources/utils.py
```diff
@@ -26,76 +31,178 @@ class TimeoutSampler:
...
+            exp: exception: Exception object to be tested
+
+        Returns:
+            bool: True if exception should be raised, False otherwise
+        """
+
+        exp_class = exp.__class__
+        if exp_class in self.exceptions_dict:
+            exception_messages = self.exceptions_dict.get(exp_class)
+            if self._is_exception_message_matched(
```
rnetser, line 185:
```suggestion
            if not exception_messages or self._is_exception_message_matched(
```

ocp_resources/utils.py
rnetser: yes

ocp_resources/utils.py
```diff
@@ -26,76 +31,178 @@ class TimeoutSampler:
...
+                exp=exp, exception_messages=exception_messages
+            ):
+                return False
+
+        else:
+            for entry in self.exceptions_dict:
+                if isinstance(exp, entry):
```
rnetser, line 192: This is always True, right? why do we need to keep this?

ocp_resources/utils.py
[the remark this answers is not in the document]
sarahbx: This was an agreement we made at the beginning of this. Please stick with our agreed upon removal date.

ocp_resources/utils.py
sarahbx: https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r681660501

ocp_resources/utils.py
sarahbx: Not always True.
```
>>> entry = ValueError()
>>> entry.__class__ == Exception
False
>>> isinstance(entry, Exception)
True
```

ocp_resources/utils.py
sarahbx: This message logic should remain within `_is_exception_message_matched()` as I have said previously to keep the logic of this method simple.

ocp_resources/utils.py
sarahbx: https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r681664442

ocp_resources/utils.py
sarahbx: how? the exception has been processed.

ocp_resources/utils.py
sarahbx: https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r678426527

ocp_resources/utils.py
sarahbx: I can expand on my example, but the short version is, `Exception` is only a default, and may not appear during runtime.

sarahbx: https://github.com/RedHatQE/openshift-python-wrapper/pull/24#discussion_r678349850

ocp_resources/utils.py
```diff
@@ -25,76 +30,183 @@ class TimeoutSampler:
...
+        An exception specifically referenced in exceptions_dict
+        OR
+        A child exception inherited from an entry in exceptions_dict
+
+        Args:
+            exp: exception: Exception object to be tested
+
+        Returns:
+            bool: True if exception should be raised, False otherwise
+        """
```
sarahbx, line 184: I've updated the docstring. We can do this as a single `for entry in dict`, which would cover all cases, if we're willing to drop the `if exp_class in dict` block.

sarahbx: https://github.com/RedHatQE/openshift-python-wrapper/commit/71246eb5557bc6bc797a34f24c1c73cd9e3bda6f

.github/workflows/code-check.yml
```diff
@@ -31,3 +31,4 @@ jobs:
     - name: Run tox
       run: |
        tox -e code-check
```
myakove, line 33: can be sone in one line
tox -e code-check,unittests

ocp_resources/utils.py
```diff
@@ -25,76 +30,219 @@ class TimeoutSampler:
...
+            exception_messages (list): Either an empty list allowing all text,
+                                       or a list of allowed strings to match against the exception text.
 
-        log = "{timeout}\n{func_log}\n{last_exception_log}".format(
-            timeout=self.wait_timeout,
-            func_log=self._func_log,
-            last_exception_log=last_exception_log,
-        )
+        Returns:
+            bool: True if exception text is allowed, False otherwise
```
myakove, line 202: True if exception text is allowed or not exception text given (in this case the exception as allowed with any massage)

ocp_resources/utils.py
```diff
@@ -25,76 +30,219 @@ class TimeoutSampler:
...
+        """
+        if not exception_messages:
+            return True
+
+        for msg in exception_messages:
+            if msg and msg in str(exp):
```
myakove, line 208: `if msg` is needed here?
you iterate over the list so `msg` exists.

myakove (commented):
few small comments, otherwise looks good.

ibesso-rh:
@sarahbx , could you please process/absorb the final result following all the comments i got on my PR (strictly docstring PR)?
https://github.com/RedHatQE/openshift-python-wrapper/pull/37

I'm closing that one now.

ocp_resources/utils.py
sarahbx: Added a comment to clarify the intention behind it.

myakove (approved):

PR #34, opened 2021-07-07
Size labeler

myakove:


sarahbx:
Looks like we're encountering the same issue:
https://github.com/CodelyTV/pr-size-labeler/issues/8
The script should also error, but it's not.

PR #35, opened 2021-07-09
Fixed an issue with high-performance templates

obaireys:
The previous attribute name HIGH_PERFORMANCE caused an AttributeError.

### DIFF
--- a/ocp_resources/template.py
+++ b/ocp_resources/template.py
@@ -15,7 +15,7 @@ class Template(NamespacedResource):
 
     class Workload:
         DESKTOP = "desktop"
-        HIGH_PERFORMANCE = "highperformance"
+        HIGHPERFORMANCE = "highperformance"
         SERVER = "server"
 
     class Flavor:


myakove:
HIGH_PERFORMANCE it's just the name of the arg, how come it fail something?

obaireys:
> HIGH_PERFORMANCE it's just the name of the arg, how come it fail something?

The high-performance templates' label is `workload.template.kubevirt.io/highperformance`, so the workload attribute's value must be `highperformance`. On line 66 (in the function `generate_template_labels`) the uppercase of the workload name is assumed to be the attribute name, at which point its value is appended to the workload label. Thus, with the previous attribute name of `HIGH_PERFORMANCE`, an AttributeError is thrown any time the `highperformance` workload is used.

rnetser:
Got it

/lgtm

rnetser:
Note that you need to add a PR to fix any dependant tests code

obaireys:
> Note that you need to add a PR to fix any dependant tests code

Noted, will do.

PR #45, opened 2021-07-18
Migration ToolKit For Virtualization Common Module 

amastbau:
##### Short description:
Common Module requires by  all MTV Resouce Handlers

### DIFF
--- a/ocp_resources/mtv.py
+++ b/ocp_resources/mtv.py
@@ -88,23 +88,13 @@ class MTV:
                 )
                 for condition in current_conditions:
                     last_condition = condition
-                    if (
-                        condition_status == condition.status
-                        and condition_type == condition.type
-                    ):
-                        if (
-                            condition_message == condition.message
-                            or condition_message is None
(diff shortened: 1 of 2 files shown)


myakove (approved):

ocp_resources/mtv.py
```diff
@@ -0,0 +1,113 @@
...
+        try:
+            for sample in samples:
+                current_conditions = (
+                    sample.items[0].status.get("conditions")
+                    if sample.items and sample.items[0].status
+                    else []
+                )
+                for condition in current_conditions:
+                    last_condition = condition
+                    if (
```
sarahbx, line 91: Would it be possible to flatten out this logic, save the conditions to boolean vars? It would be much easier to read.

sarahbx (changes requested):

ocp_resources/mtv.py
amastbau: wyt now? @sarahbx

PR #51, opened 2021-07-19
Storage: update priority class

dalia-frank:
##### Short description:
Update PriorityClass

### DIFF
--- a/ocp_resources/priority_class.py
+++ b/ocp_resources/priority_class.py
@@ -37,4 +37,4 @@ class PriorityClass(Resource):
             res["description"] = self.description
         if self.preemption_policy:
             res["preemptionPolicy"] = self.preemption_policy
-        return res    
+        return res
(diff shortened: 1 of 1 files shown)


ocp_resources/priority_class.py
```diff
@@ -7,3 +7,34 @@ class PriorityClass(Resource):
...
+            teardown=teardown,
+        )
+        self.value = value
+        self.global_default = global_default
+        self.description = description
+        self.preemption_policy = preemption_policy
+
+    def to_dict(self):
+        res = super().to_dict()
+        res.update({"value": self.value})
```
myakove, line 33: Do you want to update value even when it's None? (this is the default)

dalia-frank: Value should be a mandatory field, I updated it to None because when 'get' called, it use the init of the requested resource

dalia-frank: I was sure this is a mandatory field, but seems like if not chosen the default is 0
I'll update (all fields must be None beside name because when 'get' called, it use the init of the requested resource)

myakove (approved):

myakove (approved):

PR #53, opened 2021-07-21
Add unit test for TimeoutSampler _process_execution

sarahbx:
Add unit test to validate `_process_execution()` in `TImeoutSampler()`.

This is to provide a baseline before we continue with PR: https://github.com/RedHatQE/openshift-python-wrapper/pull/24

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -37,6 +37,7 @@ fcfn_exclude_functions =
     func,
     LOGGER,
     findall,
+    raises,
 
 enable-extensions =
     FCFN,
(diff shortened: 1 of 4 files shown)


sarahbx:
@myakove @rnetser please verify the expected functionality of `init_keyerror_raise_valueerror_with_no_msg`.

In #24 the runtime exception is expected to be raised, which I believe is the intended behavior.

sarahbx:
all testing will be in #24

PR #56, opened 2021-07-24
use self.timeout instead of timeout arg and support timeout in ProjectRequest __init__

ibesso-rh:
resource.py
- remove the timeout arg from all relevant methods
- use self.timeout when creating the TimeoutSampler

project.py
- add timeout arg to ProjectRequest.__init__

##### Short description:
1. in Resource methods, use the self.timeout instead of the timeout arg passed to the methods.
2. add the possibility to set timeout for ProjectRequest object upon creation.

##### More details:
1. timeout should be set in __init__ and when calling the methods, they should use the object's timeout attribute and not an arg.
2. I could not create a ProjectRequest object with custom timeout.

This PR fixes the wrong usage of timeout:
1. self.timeout was initialized in __init__ but never used.
2. each method would pass a timeout arg.
3. no option to override it from outside the ocp_resources.
4. no option to set a custom timeout for the ProjectRequest class.

no. 3 and 4 above.

I went over each Resource.delete() call to make sure that no timeout arg was used, and actually there was no such usage.

### DIFF
--- a/ocp_resources/daemonset.py
+++ b/ocp_resources/daemonset.py
@@ -48,13 +48,12 @@ class DaemonSet(NamespacedResource):
                 ):
                     return
 
-    def delete(self, wait=False, timeout=TIMEOUT):
+    def delete(self, wait=False):
         """
         Delete Daemonset
 
         Args:
             wait (bool): True to wait for Daemonset to be deleted.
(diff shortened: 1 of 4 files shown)


ibesso-rh:
verified the PR with automation tests.

PR #74, opened 2021-07-29
Remove nudge_delete

rnetser:
##### Short description:
Remove nudge_delete functionality

##### More details:
Upstream bug: https://github.com/kubernetes/kubernetes/issues/60807 is closed, there is no need to use nudge_delete anymore.

##### Release notes:

```release-note
Remove nudge_delete functionality; as https://github.com/kubernetes/kubernetes/issues/60807 is fixed, there's no need to forcefully clean up resources.
```


myakove:
Docs shouldn't be part of this PR, docs should be generated when there is a new release.

PR #82, opened 2021-08-03
CDI resource is not Namespace resource

myakove:
##### Short description:
CDI resource is not Namespace resource and should inherit from Resource.

##### Release notes:

```release-note
None
```

### DIFF
--- a/ocp_resources/cdi.py
+++ b/ocp_resources/cdi.py
@@ -1,13 +1,13 @@
-from ocp_resources.resource import NamespacedResource
+from ocp_resources.resource import Resource
 
 
-class CDI(NamespacedResource):
+class CDI(Resource):
     """
     CDI object.
     """
 
-    api_group = NamespacedResource.ApiGroup.CDI_KUBEVIRT_IO
+    api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
 
-    class Status(NamespacedResource.Status):
+    class Status(Resource.Status):
         DEPLOYING = "Deploying"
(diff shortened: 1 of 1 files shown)


ocp_resources/cdi.py
```diff
@@ -1,13 +1,13 @@
...
 
 
-class CDI(NamespacedResource):
+class CDI(Resource):
     """
     CDI object.
     """
 
-    api_group = NamespacedResource.ApiGroup.CDI_KUBEVIRT_IO
+    api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
```
sarahbx, line 9: This should be updated across the wrapper, there are additional places this `ApiGroup` is referenced as a `NamespacedResource`

sarahbx (changes requested):

ocp_resources/cdi.py
myakove: @sarahbx what do you mean?
each resource call his parent.

sarahbx: It does due to inheritance, but this is confusing for cluster scoped resources, using both to refer to a non-namespace scoped resource.

myakove:
```
cdi = list(CDI.get(dyn_client=admin_client))
[<ocp_resources.cdi.CDI object at 0x7f21c03f4160>]
cdi[0].name
'cdi-kubevirt-hyperconverged'
```

ibesso-rh:
/approve
ran tests that failed before.

Issue #98, opened 2021-08-09
NNCP: break sampler if reason is `NoMatchingNode`


myakove:
```
Status:
  Conditions:
    Message:               Policy does not match any node
    Reason:                NoMatchingNode
    Status:                True
    Type:                  Available 
```

When NNCP reason is `NoMatchingNode` there is no point to wait, we need to fail it.

PR #102, opened 2021-08-11
NNCP: Fail when reason is NoMatchingNode

myakove:
No need to wait 480 seconds if the reason is NoMatchingNode

#98

### DIFF
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -302,6 +302,8 @@ class NodeNetworkConfigurationPolicy(Resource):
                 return condition["reason"]
 
     def wait_for_status_success(self):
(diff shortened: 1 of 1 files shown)


ocp_resources/node_network_configuration_policy.py
```diff
@@ -311,6 +312,11 @@ def wait_for_status_success(self):
                     LOGGER.info("NNCP configured Successfully")
                     return sample
 
+                if sample == self.Conditions.Reason.NO_MATCHING_NODE:
```
rnetser, line 315: `self.Conditions.Reason.NO_MATCHING_NODE` save and re-use

rnetser (approved):

rnetser (approved):

rnetser (approved):

PR #104, opened 2021-08-12
k8s common resource labels

AdiZav:
##### Short description:
Common k8s resource labes
##### More details:
Something to consider if we end up adding functionality like adding/getting/removing resource labels -
Most resources have only some of these labels if any at all

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -223,6 +223,15 @@ class Resource(object):
     api_version = None
     singular_name = None
 
+    class Label:
+        NAME = "app.kubernetes.io/name"
+        INSTANCE = "app.kubernetes.io/instance"
+        VERSION = "app.kubernetes.io/version"
+        COMPONENT = "app.kubernetes.io/component"
+        PART_OF = "app.kubernetes.io/part-of"
+        MANAGED_BY = "app.kubernetes.io/managed-by"
+        CREATED_BY = "app.kubernetes.io/created-by"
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -223,6 +223,15 @@ class Resource(object):
     api_version = None
     singular_name = None
 
+    class Label:
+        NAME = "app.kubernetes.io/name"
```
myakove, line 227: I think what @rnetser wanted is to have
`app.kubernetes.io` under `class ApiGroup`

rnetser: Yes (i.e add "app.kubernetes.io" under  ApiGroup and add Label.NAME)

AdiZav: So when used in cnv-tests it should look like:
`    preallocation_requested_annotation = (
        f"{NamespacedResource.ApiGroup.CDI_KUBEVIRT_IO}/{preallocation_requested}"
    )`?

myakove (changes requested):
Set in class ApiGroup

ocp_resources/resource.py
```diff
@@ -223,6 +223,15 @@ class Resource(object):
     api_version = None
     singular_name = None
 
+    class Label:
```
myakove, line 226: class label can be removed.

myakove (changes requested):

PR #108, opened 2021-08-16
Add ServiceMeshMemberRoll resource for Service Mesh usage.

ofirnash1:
##### Short description:
Add ServiceMeshMemberRoll resource for Service Mesh.

##### More details:
Add SMMR resource with ApGroup: maistra.io.

Service Mesh

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -279,6 +279,7 @@ class Resource(object):
         LITMUS_IO = "litmuschaos.io"
         MACHINE_OPENSHIFT_IO = "machine.openshift.io"
         MACHINECONFIGURATION_OPENSHIFT_IO = "machineconfiguration.openshift.io"
+        MAISTRA_IO = "maistra.io"
         MONITORING_COREOS_COM = "monitoring.coreos.com"
         NETWORKADDONSOPERATOR_NETWORK_KUBEVIRT_IO = (
             "networkaddonsoperator.network.kubevirt.io"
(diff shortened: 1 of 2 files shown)


ocp_resources/ServiceMeshMemberRoll.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
```
rnetser, line 1: Please rename the file to keep the same convention as other resources.

ocp_resources/resource.py
```diff
@@ -279,6 +279,7 @@ class ApiGroup:
         LITMUS_IO = "litmuschaos.io"
         MACHINE_OPENSHIFT_IO = "machine.openshift.io"
         MACHINECONFIGURATION_OPENSHIFT_IO = "machineconfiguration.openshift.io"
+        MAISTRA_ISTIO_IO = "maistra.io"
```
rnetser, line 282: Either rename the variable or update the API group string; should match one another, so either:
MAISTRA_ISTIO_IO = "maistra.istio.io"
or
MAISTRA_IO = "maistra.io"

rnetser (changes requested):

ocp_resources/resource.py
ofirnash1: Changed to: MAISTRA_IO = "maistra.io"

ocp_resources/ServiceMeshMemberRoll.py
ofirnash1: Missed it, updated. Thanks:)

myakove (approved):

PR #114, opened 2021-08-21
add infrastructure resource

GuyAfik:
##### Short description:

add the infrastructure resource 

##### More details:

Infrastructure reference:

https://docs.openshift.com/container-platform/4.8/rest_api/config_apis/infrastructure-config-openshift-io-v1.html

simply to add the infrastructure resource of OpenShift.

In OcpOnRHV project we need to use some of its fields for some fields of the machine-set when creating it.

### DIFF
--- a/ocp_resources/infrastructure.py
+++ b/ocp_resources/infrastructure.py
@@ -1,0 +1,13 @@
+from ocp_resources.resource import Resource
+
+
+class Infrastructure(Resource):
+    """
+    Infrastructure object.
+    """
+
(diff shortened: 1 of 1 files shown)


myakove:
Already exist https://github.com/RedHatQE/openshift-python-wrapper/blob/master/ocp_resources/infrastructure.py

GuyAfik:
> Already exist https://github.com/RedHatQE/openshift-python-wrapper/blob/master/ocp_resources/infrastructure.py

ack, closing the PR, thanks

PR #118, opened 2021-08-23
ResourceEditor: ResourceField sanitizing

obaireys:
Added a static method to ResourceEditor which recursively turns
ResourceField objects into dicts to avoid issues when applying patches
created by appending/changing existing resource fields.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -13,6 +13,7 @@ from openshift.dynamic.exceptions import (
     NotFoundError,
     ServerTimeoutError,
 )
+from openshift.dynamic.resource import ResourceField
 from urllib3.exceptions import ProtocolError
 
 from ocp_resources.utils import TimeoutExpiredError, TimeoutSampler
@@ -882,7 +883,7 @@ class ResourceEditor(object):
          using an unprivileged_user; use default_client or similar instead.***
         """
 
-        self._patches = patches
+        self._patches = self._dictify_resourcefield(res=patches)
         self.action = action
         self.user_backups = user_backups
         self._backups = {}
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -969,6 +970,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         # restore backups
         self.restore()
 
+    @staticmethod
+    def _dictify_resourcefield(res):
+        """Recursively turns any ResourceField objects into dicts to avoid issues caused by appending lists, etc."""
+        if issubclass(type(res), ResourceField):
+            return ResourceEditor._dictify_resourcefield(res=dict(res.items()))
+        elif type(res) == dict:
```
rnetser, line 978:
```suggestion
        elif isinstance(res, dict):
```

ocp_resources/resource.py
```diff
@@ -969,6 +970,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            return {
+                ResourceEditor._dictify_resourcefield(
+                    res=key
+                ): ResourceEditor._dictify_resourcefield(res=value)
+                for key, value in res.items()
+            }
+        elif type(res) == list:
```
rnetser, line 985: use isinstance

ocp_resources/resource.py
```diff
@@ -969,6 +970,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            return [ResourceEditor._dictify_resourcefield(res=x) for x in res]
+        else:
```
rnetser, line 987: The `else` is redundant; can just return res

rnetser (changes requested):

ocp_resources/resource.py
```diff
@@ -969,6 +970,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         # restore backups
         self.restore()
 
+    @staticmethod
+    def _dictify_resourcefield(res):
+        """Recursively turns any ResourceField objects into dicts to avoid issues caused by appending lists, etc."""
+        if issubclass(type(res), ResourceField):
```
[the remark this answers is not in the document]
obaireys: I would prefer to cover the case where some subclass of `ResourceField` for some specific resource type, as opposed to the primitives `list` and `dict` whose use is independent of `ResourceField`.

rnetser (approved):

PR #124, opened 2021-08-25
Add vm status

rnetser:
##### Short description:
Add VM Status.STARTING and move Status.RUNNING to resource

### DIFF
--- a/ocp_resources/virtual_machine.py
+++ b/ocp_resources/virtual_machine.py
@@ -26,6 +26,9 @@ class VirtualMachine(NamespacedResource):
         ALWAYS = "Always"
         RERUNONFAILURE = "RerunOnFailure"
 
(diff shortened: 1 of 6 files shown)


sarahbx:
👍 lgtm

ocp_resources/virtual_machine.py
```diff
@@ -26,6 +26,14 @@ class RunStrategy:
         ALWAYS = "Always"
         RERUNONFAILURE = "RerunOnFailure"
 
+    class Status(NamespacedResource.Status):
```
myakove, line 29: All the statuses should be under resource

rnetser: If you have a status that's specific for a card I do not think it should go under resource

PR #131, opened 2021-08-30
resource.delete: accept body

myakove:
In some cases we need to send body to delete.
for exmaple:
body=kubernetes.client.V1DeleteOptions(grace_period_seconds=0)


akalenyu:
I see what you mean, we will need to change daemonset as well?
https://github.com/RedHatQE/openshift-python-wrapper/blob/master/ocp_resources/daemonset.py#L66

PR #133, opened 2021-08-30
Convert self.api() to property

myakove:
##### More details:
We use self.api without **kwargs in our code, it makes sense to convert it to property.
Add full_api method to maintain its functionality.


sarahbx:
Would you be able to add a PR description please? 🙂

sarahbx (approved):
lgtm 👍

PR #138, opened 2021-08-31
add ingress resource

ibesso-rh:
##### Short description:
add a new resource: ingress

disconnected environment testing.
this resource allows access to the internal cluster domain.


ocp_resources/ingress.py
```diff
@@ -0,0 +1,5 @@
+from ocp_resources.resource import Resource
+
+
+class Ingress(Resource):
```
sarahbx, line 4: Ingress is a NamespacedResource

sarahbx (changes requested):

ocp_resources/ingress.py
```diff
@@ -0,0 +1,5 @@
...
+    api_group = Resource.ApiGroup.CONFIG_OPENSHIFT_IO
```
sarahbx, line 5: IIRC, this is also under the networking api group

myakove: @sarahbx I guess you are right, here is an example I found
```apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: hello-openshift
spec:
  rules:
  - host: hello-openshift.yourcluster.example.com
    http:
      paths:
      - backend:
          # Forward to a Service called 'hello-openshift'
          service:
            name: hello-openshift
            port:
              number: 8080
        path: /
        pathType: Exact
```

Don't think it's namespaced resource, but it should be use networking api_group.

/unapprove

sarahbx: We can get list/watch without the namespace, which returns an IngressList, but creation/deletion is namespaced
https://docs.openshift.com/container-platform/4.8/rest_api/network_apis/ingress-networking-k8s-io-v1.html

myakove: BUT

ipdb> resource_definitions_dict["Ingress"]
{'namespaced': False, 'group': 'config.openshift.io'}


I'm confused.

rnetser: This is what got me confused but reading again the docs, yes - it should be namespaced.

sarahbx: config api does also have access to ingress, but I think it should be separate from the namespaced ingress

https://docs.openshift.com/container-platform/4.8/rest_api/config_apis/ingress-config-openshift-io-v1.html

rnetser:
@sarahbx - from the crd:

```  names:
    kind: Ingress
    listKind: IngressList
    plural: ingresses
    singular: ingress
  scope: Cluster
```

rnetser:
i did see examples where it can be created under a ns, so what are we missing?

sarahbx:
Perhaps we should handle this as `ClusterIngress` and `Ingress`?
```
$ oc api-resources | grep 'Ingress$'
NAME                                  SHORTNAMES                             APIVERSION                                     NAMESPACED   KIND
ingresses                                                                    config.openshift.io/v1                         false        Ingress
ingresses                             ing                                    networking.k8s.io/v1                           true         Ingress
```

ocp_resources/ingress.py
ibesso-rh: it's not from what i tried (i tried the oc command first, obviously, without namespace). if it should be changed, a bug will be in order.

i was provided the command and had to figure out by myself how to use the resource to fetch it.

ocp_resources/ingress.py
ibesso-rh: what do you mean by that?
I could not fetch it by the networking. i need the specific ingress resource.
please clarify if you refer to the code i added.

ibesso-rh:
> Perhaps we should handle this as `ClusterIngress` and `Ingress`?
> 
> ```
> $ oc api-resources | grep 'Ingress$'
> NAME                                  SHORTNAMES                             APIVERSION                                     NAMESPACED   KIND
> ingresses                                                                    config.openshift.io/v1                         false        Ingress
> ingresses                             ing                                    networking.k8s.io/v1                           true         Ingress
> ```

not sure if this is a review comment, but when i used Ingresses as the class name, it fetched nothing, which is logical (after the fact), since the kind is Ingress.

i just added those comments from my attempts to make it work.

myakove:
@ibesso-rh please rebase

ibesso-rh:
> @ibesso-rh please rebase

rebased.

ibesso-rh:
tox failed:
resources_definitions_errors = ['Resource Ingress should be namespaced']

    def test_resources_definitions(resources_definitions_errors):
>       assert not resources_definitions_errors, "\n".join(resources_definitions_errors)
E       AssertionError: Resource Ingress should be namespaced
E       assert not ['Resource Ingress should be namespaced']

tests/test_validate_resources.py:123: AssertionError


on latest cnv build i can still access without namespace:

$ oc get ingresses.config.openshift.io cluster
NAME      AGE
cluster   14h

therefore, we have a conflict between the checks here and the product.

myakove:
@ibesso-rh please rebase

ibesso-rh:
/assign @myakove 

the 3 CI checks fail on the same error:
```
Building wheel for pygit2 (PEP 517): started
    Building wheel for pygit2 (PEP 517): finished with status 'error'
    ERROR: Command errored out with exit status 1:
   
```

PR #140, opened 2021-09-01
dv.wait - raise TimeoutExpiredError if DV's status is Pending

rnetser:
##### Short description:
Avoid waiting for a timeout for a DV to reach Succeeded status if it reached Pending within a given timeout

### DIFF
--- a/ocp_resources/datavolume.py
+++ b/ocp_resources/datavolume.py
@@ -4,7 +4,7 @@ import logging
 
 from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
 from ocp_resources.resource import TIMEOUT, NamespacedResource, Resource
-from ocp_resources.utils import TimeoutSampler
+from ocp_resources.utils import TimeoutExpiredError, TimeoutSampler
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


ocp_resources/datavolume.py
```diff
@@ -183,7 +184,17 @@ def wait_deleted(self, timeout=TIMEOUT):
...
         return self.pvc.wait_deleted(timeout=timeout)
 
-    def wait(self, timeout=600):
+    def wait(self, timeout=600, failure_timeout=30):
+        # Avoid waiting for "Succeeded" status if DV's in pending/failed status
+        for sample in TimeoutSampler(
+            wait_timeout=failure_timeout,
+            sleep=1,
+            func=lambda: self.instance.get("items", [""])[0].status
+            in [self.Status.PENDING, self.Status.FAILED],
```
sarahbx, line 193: Could we move this to its own method?
`self.wait_for_status_not(status=[self.Status.PENDING, self.Status.FAILED], timeout=failure_timeout)`

sarahbx (changes requested):

ocp_resources/datavolume.py
rnetser: done

PR #147, opened 2021-09-02
Add ConflictError to DEFAULT_CLUSTER_RETRY_EXCEPTIONS

dbasunag:
Without this change currently, we are having to use TimeoutSampler and add ConflictError to exceptions_dict, every time we make calls to ResourceEditor
##### Bug:
None

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -26,7 +26,6 @@ DEFAULT_CLUSTER_RETRY_EXCEPTIONS = {
     ConnectionResetError: [],
     InternalServerError: ["etcdserver: leader changed"],
     ServerTimeoutError: [],
-    ConflictError: [],
 }
 
 LOGGER = logging.getLogger(__name__)
@@ -656,12 +655,14 @@ class Resource:
         self.api.replace(body=resource_dict, name=self.name, namespace=self.namespace)
 
     @staticmethod
-    def _retry_cluster_exceptions(func):
+    def _retry_cluster_exceptions(
+        func, exception_dict=DEFAULT_CLUSTER_RETRY_EXCEPTIONS
+    ):
         sampler = TimeoutSampler(
             wait_timeout=10,
             sleep=1,
             func=func,
-            exceptions_dict=DEFAULT_CLUSTER_RETRY_EXCEPTIONS,
+            exceptions_dict=exception_dict,
             print_log=False,
         )
         for sample in sampler:
@@ -1091,4 +1092,8 @@ class ResourceEditor:
                         resource_dict=patch
                     )  # replace the resource metadata
 
-        return Resource._retry_cluster_exceptions(func=_patch)
+        exceptions_dict = {ConflictError: []}
+        exceptions_dict.update(DEFAULT_CLUSTER_RETRY_EXCEPTIONS)
+        return Resource._retry_cluster_exceptions(
+            func=_patch, exception_dict=exceptions_dict
+        )
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -656,14 +656,20 @@ class Resource:
 
     @staticmethod
     def _retry_cluster_exceptions(
-        func, exception_dict=DEFAULT_CLUSTER_RETRY_EXCEPTIONS
+        func,
+        **kwargs,
     ):
+        exceptions_dict = DEFAULT_CLUSTER_RETRY_EXCEPTIONS
+        if "exceptions_dict" in kwargs:
+            exceptions_dict.update(kwargs["exceptions_dict"])
+            del kwargs["exceptions_dict"]
         sampler = TimeoutSampler(
             wait_timeout=10,
             sleep=1,
             func=func,
-            exceptions_dict=exception_dict,
             print_log=False,
+            exceptions_dict=exceptions_dict,
+            **kwargs,
         )
         for sample in sampler:
             return sample
@@ -968,12 +974,12 @@ class ResourceEditor:
         }
 
         # apply changes
-        self._apply_patches(
+        self._apply_patches_sampler(
             patches=patches_to_apply, action_text="Updating", action=self.action
         )
 
     def restore(self):
-        self._apply_patches(
+        self._apply_patches_sampler(
             patches=self._backups, action_text="Restoring", action=self.action
         )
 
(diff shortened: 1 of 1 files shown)


rnetser:
Do you plan on wrapping `_apply_patches`  the same `Resource._retry_cluster_exceptions(func=_get)` is done in Resource.get()?

sarahbx:
> Do you plan on wrapping `_apply_patches` the same `Resource._retry_cluster_exceptions(func=_get)` is done in Resource.get()?

+1, we should localize this to ResourceEditor

myakove:
@dbasunag  Small advice, don't work on your master branch. Create a local branch and work from there.

dbasunag:
@rnetser and @sarahbx I have updated _apply_patches as suggested.

@myakove I will create local branch going forward, this is indeed a bad idea.

ocp_resources/resource.py
```diff
@@ -25,6 +26,7 @@
     ConnectionResetError: [],
     InternalServerError: ["etcdserver: leader changed"],
     ServerTimeoutError: [],
+    ConflictError: [],
```
sarahbx, line 29: Any `ConflictError` exception handling will need to be limited to within `ResourceEditor`, it should not be added to the defaults for all resources.

sarahbx (changes requested):

ocp_resources/resource.py
dbasunag: Ok, I will update `_retry_cluster_exceptions` to accept `exceptions_dict` as a parameter, to achieve this.

ocp_resources/resource.py
```diff
@@ -1053,36 +1056,44 @@ def _apply_patches(patches, action_text, action):
...
-                patch["metadata"] = {}
 
-            # the api requires this field to be present in a yaml patch for
-            # some resource kinds even if it is not changed
-            if "name" not in patch["metadata"]:
-                patch["metadata"]["name"] = resource.name
-
-            if action == "update":
-                resource.update(resource_dict=patch)  # update the resource
+        def _patch():
```
myakove, line 1060: This is not needed.
Just create a function `_apply_patches_sampler` and call it instead of `_apply_patches`.

```python
def _apply_patches_sampler(patches, action_text, action):
    return Resource._retry_cluster_exceptions(
            func=_apply_patches, exception_dict=exceptions_dict, patches=patches, action_text=action_text, action=action
        )

you will need to add kwargs to `_retry_cluster_exceptions`
```

myakove (changes requested):

ocp_resources/resource.py
dbasunag: Ok, I will implement it this way.

ocp_resources/resource.py
```diff
@@ -654,13 +655,21 @@ def update_replace(self, resource_dict):
         self.api.replace(body=resource_dict, name=self.name, namespace=self.namespace)
 
     @staticmethod
-    def _retry_cluster_exceptions(func):
+    def _retry_cluster_exceptions(
+        func,
```
myakove, line 659: Any reason why you have here multiple lines?

ocp_resources/resource.py
```diff
@@ -654,13 +655,21 @@ def update_replace(self, resource_dict):
...
+        **kwargs,
+    ):
+        exceptions_dict = DEFAULT_CLUSTER_RETRY_EXCEPTIONS
+        if "exceptions_dict" in kwargs:
```
myakove, line 663: `exceptions_dict` should be in arg, kwargs if for the func kwargs to pass to the sampler.

myakove (changes requested):

ocp_resources/resource.py
dbasunag: I thought this was done by "pre-commit", I did not create it like this intentionally, let me try to see if I can undo this.

ocp_resources/resource.py
dbasunag: Ok, that makes sense. I will update this PR soon.

ocp_resources/resource.py
```diff
@@ -654,13 +655,21 @@ def update_replace(self, resource_dict):
         self.api.replace(body=resource_dict, name=self.name, namespace=self.namespace)
 
     @staticmethod
-    def _retry_cluster_exceptions(func):
+    def _retry_cluster_exceptions(
+        func,
+        **kwargs,
```
rnetser, line 660: Please replace with a parameter with a default; we should avoid using kwargs - 1. for usability 2. we know what we expect
so something like:
`cluster_exceptions=DEFAULT_CLUSTER_RETRY_EXCEPTIONS`

A user can then either not pass anything or pass their own (where they can decide which expections they's like to use)

ocp_resources/resource.py
```diff
@@ -654,13 +655,21 @@ def update_replace(self, resource_dict):
         self.api.replace(body=resource_dict, name=self.name, namespace=self.namespace)
 
     @staticmethod
-    def _retry_cluster_exceptions(func):
+    def _retry_cluster_exceptions(
```
rnetser, line 658:
```suggestion
    def retry_cluster_exceptions(
```
this is no longer a private function

rnetser (changes requested):

ocp_resources/resource.py
dbasunag: @rnetser this change was done based on @myakove 's recommendation on a previous review comment. I originally had something like what you recommended. Can you both please advice on how you want me to proceed here? https://github.com/RedHatQE/openshift-python-wrapper/pull/147#discussion_r701690506

ocp_resources/resource.py
dbasunag: ok. I will update it and ensure the associated calls are changed as well.

dbasunag:
@myakove / @rnetser please let me know how you want me to proceed here.

ocp_resources/resource.py
myakove, line 660: What I seggested is:
```python
def _retry_cluster_exceptions(func, exceptions_dict, **kwargs)
```
func and exceptions_dict belong to this function while kwargs pass to TimeoutSampler as function kwargs.

@rnetser

myakove (changes requested):

ocp_resources/resource.py
dbasunag: ok, I misunderstood then. I will address the comment from @rnetser accordingly. Thanks for the clarification @myakove .

ocp_resources/resource.py
```diff
@@ -1086,3 +1092,15 @@ def _apply_patches(patches, action_text, action):
                 resource.update_replace(
                     resource_dict=patch
                 )  # replace the resource metadata
+
+    @staticmethod
```
myakove, line 1096: `@staticmethod` should be removed and `_apply_patches` should be called with self.
`func=self._apply_patches`

myakove (changes requested):

ocp_resources/resource.py
dbasunag: ok. I will update this

ocp_resources/resource.py
```diff
@@ -1086,3 +1092,14 @@ def _apply_patches(patches, action_text, action):
...
                     resource_dict=patch
                 )  # replace the resource metadata
+
+    def _apply_patches_sampler(self, patches, action_text, action):
+        exceptions_dict = {ConflictError: []}
+        exceptions_dict.update(DEFAULT_CLUSTER_RETRY_EXCEPTIONS)
+        return Resource.retry_cluster_exceptions(
+            func=self._apply_patches,
+            exceptions_dict=exceptions_dict,
+            patches=patches,
```
sarahbx, line 1102: The `patches` variable will be modified by _apply_patches, please add a .copy() unless this is intentional.

dbasunag: @sarahbx This update seems intentional. I also want to add, my PR is not changing anything in  `_apply_patches `, so If this comment needs to be addressed, it should be addressed in a separate PR.

With TimeoutSampler changes taken out of IUO area, without this piece to handle conflict error, we can have significant fail rate in current release testing.

PR #149, opened 2021-09-02
Add hostname property to Node

Anatw:
##### Short description:
Will be used when specifying a node_selector.

In AWS clusters, when specifying a node_selector, the label kubernetes.io/hostname should be used.


sarahbx (approved):
lgtm

PR #168, opened 2021-09-05
host resource

amastbau:
##### Short description:
openshift mtv Host CR

### DIFF
--- a/ocp_resources/host.py
+++ b/ocp_resources/host.py
@@ -1,0 +1,64 @@
+from ocp_resources.mtv import MTV
+from ocp_resources.resource import NamespacedResource
+
+
+class Host(NamespacedResource, MTV):
+    """
+    Migration Toolkit For Virtualization (MTV) Host object.
+    """
+
+    api_group = NamespacedResource.ApiGroup.FORKLIFT_KONVEYOR_IO
+
+    def __init__(
+            self,
+            name=None,
+            namespace=None,
+            host_id=None,
+            ip_address=None,
+            provider_name=None,
+            provider_namespace=None,
+            secret_name=None,
+            secret_namespace=None,
+            client=None,
+            teardown=True,
+            yaml_file=None,
+    ):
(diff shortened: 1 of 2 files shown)


ocp_resources/host.py
```diff
@@ -0,0 +1,63 @@
+from ocp_resources.mtv import MTV
+from ocp_resources.resource import NamespacedResource
+
+
+class Host(NamespacedResource, MTV):
+    """
+    Migration Toolkit For Virtualization (MTV) Host object.
```
rnetser, line 7:
```suggestion
    Migration Toolkit For Virtualization (MTV) Host resource.
```

rnetser (changes requested):

ocp_resources/host.py
```diff
@@ -0,0 +1,63 @@
...
+            teardown=teardown,
+            yaml_file=yaml_file,
+        )
+        self.host_id = host_id
+        self.ip_address = ip_address
+        self.provider_name = provider_name
+        self.provider_namespace = provider_namespace
+        self.secret_name = secret_name or f"{self.name}-secret"
+        self.secret_namespace = secret_namespace or self.namespace
+        self.condition_message_ready = self.ConditionMessage.HOST_READY
```
myakove, line 39: This `self.condition_message_ready` not used and I guess it's net needed, whenever you need it you can call it via the Host resource that was created

myakove (changes requested):

ocp_resources/host.py
amastbau: not sure what you mean
it is in
    def wait_for_condition_ready(self):
        self.wait_for_resource_status(
            condition_message=self.condition_message_ready,
            condition_status=self.Condition.Status.TRUE,
            condition_type=self.Condition.READY,
        )

amastbau: in mtv.py

PR #181, opened 2021-09-14
test_resources_definitions: Add exclude resource

myakove:
##### Short description:
Ingress resource is both namespace and not namespace resource (depend on the api_group)
while `oc get ingress.config.openshift.io` works `oc get inress -A` show nothing.

Exclude ingress from the test for now.

### DIFF
--- a/tests/test_validate_resources.py
+++ b/tests/test_validate_resources.py
@@ -10,6 +10,9 @@ import pytest
 import requests
 
 
+EXCLUDE_RESOURCES = {"ingress.py": "Ingress"}
+
+
 def _process_api_type(api_type, api_value, resource_dict, cls):
     if api_type == "api_group":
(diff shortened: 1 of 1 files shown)


tests/test_validate_resources.py
```diff
@@ -94,9 +97,13 @@ def resources_definitions_errors(resources_definitions):
     errors = []
     for _file in _resource_file():
         with open(_file, "r") as fd:
+            exclude_cls = EXCLUDE_RESOURCES.get(_file.split("/")[-1])
```
sarahbx, line 100: `os.path.basename(_file)`

myakove: ACK

sarahbx (approved):
👍 lgtm

PR #185, opened 2021-09-18
namespae resource: support get raw object in get()

myakove:
`label_selector` is not working when we have multiple resources with the same name
We return `openshift-python-wrapper` python object and this leads to get every time different resource when calling `.instance`

issue was opened in `openshift-restclient-python`:
https://github.com/openshift/openshift-restclient-python/issues/415

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -818,13 +818,14 @@ class NamespacedResource(Resource):
             raise ValueError("name and namespace or yaml file is required")
 
     @classmethod
-    def get(cls, dyn_client, singular_name=None, *args, **kwargs):
+    def get(cls, dyn_client, singular_name=None, raw=False, *args, **kwargs):
         """
         Get resources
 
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -818,13 +818,14 @@ def __init__(
...
     @classmethod
-    def get(cls, dyn_client, singular_name=None, *args, **kwargs):
+    def get(cls, dyn_client, singular_name=None, raw=False, *args, **kwargs):
         """
         Get resources
 
         Args:
             dyn_client (DynamicClient): Open connection to remote cluster
             singular_name (str): Resource kind (in lowercase), in use where we have multiple matches for resource
+            raw (bool): If True return rew object from openshift-restclient-python
```
dbasunag, line 828: Typo: rew

myakove:
@dbasunag please set verified if verified so we can merge it.

PR #228, opened 2021-11-03
Fixing 'cutover' position in migration mtv resource

mguetta1:


amastbau:
need to verify this doesn't break 2.1

sarahbx:
lgtm 👍

PR #230, opened 2021-11-04
Add DataImportCron and DataSource resources

sarahbx:


rnetser:
This code is not fully integrated and in the product.
Add WIP and remove the verify label until we have it in testable code.

ocp_resources/data_import_cron.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    api_group = NamespacedResource.ApiGroup.CDI_KUBEVIRT_IO
+
+    def __init__(
+        self,
+        name=None,
+        namespace=None,
+        client=None,
+        source=None,
```
rnetser, line 16: This is no longer source but 
```Template cdiv1.DataVolume `json:"template"```

https://github.com/kubevirt/containerized-data-importer/blob/main/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go#L482

rnetser (changes requested):

PR #232, opened 2021-11-07
Update mtv.py

amastbau:

### DIFF
--- a/ocp_resources/mtv.py
+++ b/ocp_resources/mtv.py
@@ -71,7 +71,7 @@ class MTV:
         condition_message=None,
         condition_reason=None,
         condition_category=None,
-        wait_timeout=600,
+        wait_timeout=1200,
     ):
         """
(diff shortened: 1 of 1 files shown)


ocp_resources/mtv.py
```diff
@@ -71,7 +71,7 @@ def wait_for_resource_status(
         condition_message=None,
         condition_reason=None,
         condition_category=None,
-        wait_timeout=600,
+        wait_timeout=1200,
```
myakove, line 74: please send 1200 when you call the function, no need to change the default.

myakove (changes requested):

ocp_resources/mtv.py
amastbau: sorry, did not mean to send this PR

amastbau:
will send another PR to pass the timeout

PR #242, opened 2021-11-15
namespace resource: support get raw object in get() - Cherry-pick

dbasunag:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_language_version:
 
 repos:
   - repo: https://github.com/hadialqattan/pycln
-    rev: "v0.0.1-beta.1"
+    rev: "v1.0.3"
     hooks:
       - id: pycln
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_language_version:
 
 repos:
   - repo: https://github.com/hadialqattan/pycln
-    rev: "v1.0.3"
+    rev: "v0.0.1-beta.1"
     hooks:
       - id: pycln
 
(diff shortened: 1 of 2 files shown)


myakove:
Rename the title, Cherry-pick is not informative title.

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ default_language_version:
 
 repos:
   - repo: https://github.com/hadialqattan/pycln
-    rev: "v0.0.1-beta.1"
+    rev: "v1.0.3"
```
myakove, line 6: This change is not related to this PR.

myakove (changes requested):

.pre-commit-config.yaml
dbasunag: This change was picked up by cherry-pick command. I simply resolved the conflict by picking what is in head. Also looks like this change is part of cherrypick against 1.3.10. https://github.com/RedHatQE/openshift-python-wrapper/commit/0b684ae7d12063b55d208670d47f85e397863023. Please suggest what needs to be done.

rnetser: When resolving a conflict and working on a different repo branch (i.e cherrypick from master to 1.5.x), you should use the code that exists in the branch that you're working on (and not the code that you've cherry-picked from).
This way you avoid adding extra code that should not be cherrypicked.

dbasunag: Ok, I will remove this, since it does not look like this is in 1.5.3.54. Looks like the same mistake happened with the last cherry-pick(https://github.com/RedHatQE/openshift-python-wrapper/commit/0b684ae7d12063b55d208670d47f85e397863023), please have someone take a look at it and undo the same as well.

dbasunag: @myakove and @rnetser , I have removed changes from .pre-commit-config.yaml

PR #247, opened 2021-11-22
Adding MTV ForkliftController resource

mguetta1:


ocp_resources/forkliftcontroller.py
```diff
@@ -0,0 +1,35 @@
+from ocp_resources.resource import NamespacedResource
+
+
+class ForkliftController(NamespacedResource):
+    """
+    Migration Toolkit For Virtualization (MTV) ForkliftController object
+    """
+
+    api_version = NamespacedResource.ApiVersion.V1BETA1
```
myakove, line 9: Do we have api_group for this resource?
Better to use it if we have, since then we can grub the api_version automatically

mguetta1: Actually we do have. I will use it, thanks.

mguetta1: Done

ocp_resources/forkliftcontroller.py
```diff
@@ -0,0 +1,35 @@
...
+    Migration Toolkit For Virtualization (MTV) ForkliftController object
+    """
+
+    api_group = NamespacedResource.ApiGroup.FORKLIFT_KONVEYOR_IO
+
+    def __init__(
+        self,
+        name=None,
+        namespace=None,
+        data=None,
```
amastbau, line 15: Is this here by accident?

amastbau (changes requested):

ocp_resources/forkliftcontroller.py
mguetta1: Yes. Thanks.
Currently, we need nothing but to get the ForkliftController resource

ocp_resources/forkliftcontroller.py
```diff
@@ -0,0 +1,10 @@
+from ocp_resources.mtv import MTV
```
rnetser, line 1: Please rename the file to match the convention (i.e underscore between words)

rnetser (changes requested):

ocp_resources/forkliftcontroller.py
mguetta1: Done

PR #249, opened 2021-11-23
Support generateName for resource name

myakove:


ocp_resources/resource.py
```diff
@@ -398,7 +406,9 @@ def _base_body(self):
         return {
             "apiVersion": self.api_version,
             "kind": self.kind,
-            "metadata": {"name": self.name},
+            "metadata": {"generateName": self.name}
```
akalenyu, line 409: mby we want `self.name` followed by a `-` so it doesn't get created as:
```bashnamespace/module-namehbckf created```

PR #299, opened 2021-12-21
log YAML output on create, update and delete

myakove:


ocp_resources/resource.py
```diff
@@ -640,6 +640,7 @@ def create(self, body=None, wait=False):
             data.update(body)
 
         LOGGER.info(f"Posting {data}")
```
yossisegev, line 642: I think that this one is redundant now.

yossisegev (changes requested):

yossisegev (commented):
-

ocp_resources/logger.py
```diff
@@ -0,0 +1,32 @@
+import logging
+import os
+
+from colorlog import ColoredFormatter
+
+
+def get_logger(name):
+    log_level = os.environ.get("OPW_LOG_LEVEL", "INFO")
```
rnetser, line 8: 1. Can you please update the README?
2. What does OPW stand for?

rnetser (changes requested):
"ceate" -> "create" in the title and description

ocp_resources/logger.py
yossisegev: > What does OPW stand for?

My guess is "Openshift Python Wrapper".

myakove: right, please suggest a better name.

rnetser: why not OCP_PYTHON_WRAPPER_LOG_LEVEL ? long but self explanatory.

rnetser (approved):

rnetser (approved):

PR #325, opened 2022-01-02
Support dry-run

myakove:
Add support to create resource with dry_run flag.

This will not create the resource, but only simulate the creation.

### DIFF
--- a/ocp_resources/virtual_machine.py
+++ b/ocp_resources/virtual_machine.py
@@ -42,6 +42,7 @@ class VirtualMachine(NamespacedResource):
         privileged_client=None,
         yaml_file=None,
         delete_timeout=TIMEOUT_4MINUTES,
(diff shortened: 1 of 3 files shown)


rnetser (approved):

rnetser (approved):

ocp_resources/virtual_machine.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         privileged_client=None,
         yaml_file=None,
         delete_timeout=TIMEOUT_4MINUTES,
+        **kwargs,
```
sarahbx, line 45: This should be added to all the classes in this PR

myakove: Right, @Anatw will cherry-pick and add them

PR #327, opened 2022-01-02
Add storage API to data volume

dalia-frank:
A new feature available in 4.10 is storage API
  that will replace the PVC API.
  At the moment, the default API of DataVolume class
  will remain PVC as both available in 4.10.

### DIFF
--- a/ocp_resources/datavolume.py
+++ b/ocp_resources/datavolume.py
@@ -75,11 +75,11 @@ class DataVolume(NamespacedResource):
         storage_class=None,
         url=None,
         content_type=ContentType.KUBEVIRT,
-        access_modes=AccessMode.RWO,
+        access_modes=None,
         cert_configmap=None,
         secret=None,
         client=None,
-        volume_mode=VolumeMode.FILE,
+        volume_mode=None,
         hostpath_node=None,
         source_pvc=None,
         source_namespace=None,
@@ -90,6 +90,7 @@ class DataVolume(NamespacedResource):
         privileged_client=None,
         yaml_file=None,
         delete_timeout=TIMEOUT_4MINUTES,
+        use_storage_api=False,
     ):
         super().__init__(
             name=name,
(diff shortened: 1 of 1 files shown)


ocp_resources/datavolume.py
```diff
@@ -106,40 +107,47 @@ def __init__(
         self.secret = secret
         self.content_type = content_type
         self.size = size
-        self.access_modes = access_modes
+        self.access_modes = (
+            access_modes if use_storage_api or access_modes else self.AccessMode.ROX
+        )
```
myakove, line 112:
``` self.access_modes = (
            access_modes if use_storage_api or access_modes else self.AccessMode.ROX
       )
```
Why this is needed? You still use access_mode in both cases.

ocp_resources/datavolume.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         privileged_client=None,
         yaml_file=None,
         delete_timeout=TIMEOUT_4MINUTES,
+        use_storage_api=False,
```
myakove, line 93:
```suggestion
        storage_api="pvc",
```
I prefer to see here `pvc` as default, and add `TODO` to switch to `storage` when needed or remove it and use only `storage`

ocp_resources/datavolume.py
```diff
@@ -106,40 +107,47 @@ def __init__(
...
         self.source_pvc = source_pvc
         self.source_namespace = source_namespace
         self.multus_annotation = multus_annotation
         self.bind_immediate_annotation = bind_immediate_annotation
         self.preallocation = preallocation
+        self.use_storage_api = use_storage_api
 
     def to_dict(self):
         res = super().to_dict()
+        api_to_use = "storage" if self.use_storage_api else "pvc"
```
myakove, line 127: Should be set after line 129, no need to save something that may never be used before needed.

myakove (changes requested):

ocp_resources/datavolume.py
dalia-frank: Ack

ocp_resources/datavolume.py
dalia-frank: Ack

ocp_resources/datavolume.py
dalia-frank: will update so the DV won't have a default value

myakove (approved):

rnetser (approved):

PR #331, opened 2022-01-04
remove validate_create

AdiZav:
func wait_for_status_success should be enough for validating nncp deployment.


myakove:
@AdiZav I think `wait_for_interface_deleted` can be removed too

AdiZav:
> @AdiZav I think `wait_for_interface_deleted` can be removed too

ACK

PR #335, opened 2022-01-04
NNCP: get NNCE for node

myakove:
##### Short description:
Get NNCE from NNCP by node.

### DIFF
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -75,7 +75,7 @@ class NodeNetworkConfigurationPolicy(Resource):
             delete_timeout=delete_timeout,
         )
         self.desired_state = {"interfaces": []}
-        self.worker_pods = worker_pods
+        self.worker_pods = worker_pods or []
         self.mtu = mtu
         self.capture = capture
         self.mtu_dict = {}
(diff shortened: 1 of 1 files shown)


ocp_resources/node_network_configuration_policy.py
```diff
@@ -440,6 +440,11 @@ def nnces(self):
                 nnces.append(nnce)
         return nnces
 
+    def node_nnce(self, node_name):
+        for nnce in self.nnces:
+            if nnce.instance.metadata.labels["nmstate.io/node"] == node_name:
```
geetikakay, line 445: maybe use : nnce.labels['nmstate.io/node'] directly

ocp_resources/node_network_configuration_policy.py
```diff
@@ -440,6 +440,11 @@ def nnces(self):
                 nnces.append(nnce)
         return nnces
 
+    def node_nnce(self, node_name):
+        for nnce in self.nnces:
```
rnetser, line 444: return [nnce for nnce in self.nnces if  nnce.labels["nmstate.io/node"] == node_name]

rnetser (changes requested):

rnetser (approved):

ocp_resources/node_network_configuration_policy.py
```diff
@@ -440,6 +440,11 @@ def nnces(self):
                 nnces.append(nnce)
         return nnces
 
+    def node_nnce(self, node_name):
+        return [
+            nnce for nnce in self.nnces if nnce.labels["nmstate.io/node"] == node_name
```
geetikakay, line 445: @myakove @rnetser I am bit confused here . why we make return output as list. if policy name is known, i should just be getting one enactment (NodeNetworkConfigurationEnactment object. )

myakove: You are right, I will fix it.

PR #349, opened 2022-01-10
Add support to dry_run flag

Anatw:
##### Short description:
Add support to create resources using the dry_run flag.

##### More details:
This will not create the resource but only simulate the creation.
Support added in all classes that inherit from Resource.


myakove:
should be done for all resources.

myakove (approved):

rnetser (approved):

Anatw:
\verified

PR #436, opened 2022-02-21
Support multiple tags in imagesteam

duyanyan:

### DIFF
--- a/ocp_resources/image_stream.py
+++ b/ocp_resources/image_stream.py
@@ -14,10 +14,8 @@ class ImageStream(NamespacedResource):
         name=None,
         namespace=None,
         client=None,
-        image_repository=None,
-        scheduled=None,
         lookup_policy=False,
-        tag=None,
+        tags=None,
         teardown=True,
         privileged_client=None,
         yaml_file=None,
@@ -34,9 +32,7 @@ class ImageStream(NamespacedResource):
             delete_timeout=delete_timeout,
             **kwargs,
         )
-        self.image_repository = image_repository
-        self.scheduled = scheduled
-        self.tag = tag
+        self.tags = tags
         self.lookup_policy = lookup_policy
 
     def to_dict(self):
@@ -44,17 +40,6 @@ class ImageStream(NamespacedResource):
         if self.yaml_file:
             return res
         res.update(
-            {
-                "spec": {
-                    "lookupPolicy": {"local": self.lookup_policy},
-                    "tags": {
-                        "from": {"kind": "DockerImage", "name": self.image_repository},
-                        "name": self.tag,
-                        "referencePolicy": {"type": "Source"},
-                    },
-                }
-            }
+            {"spec": {"lookupPolicy": {"local": self.lookup_policy}, "tags": self.tags}}
         )
-        if self.scheduled:
-            res["spec"]["tags"]["importPolicy"] = self.scheduled
         return res


ocp_resources/image_stream.py
```diff
@@ -47,14 +47,18 @@ def to_dict(self):
...
+                        {
+                            "from": {
+                                "kind": "DockerImage",
+                                "name": self.image_repository,
+                            },
+                            "importPolicy": {"scheduled": self.scheduled},
+                            "name": self.tag,
+                            "referencePolicy": {"type": "Source"},
+                        }
+                    ],
```
myakove, line 60: please do 
```
if self.scheduled is not None:
            res["spec"]["tags"]["importPolicy"] = self.scheduled
```

myakove (changes requested):

ocp_resources/image_stream.py
duyanyan: I'm afraid we have to set it in default false, I missed ["scheduled"] which will cause schedule not working well:
KeyError: 'importPolicy'

myakove: Must we have `scheduled` under tags always?

duyanyan: scheduled is not a must value if we don't need it, but when we set it to false, it is same as without it like:
`      importPolicy: {} `

myakove (changes requested):

ocp_resources/image_stream.py
duyanyan: Fixed. Please review again.

rnetser:
```
Also tags may have multiple tags but we limit the user to set only one?
```
?

duyanyan: Updated the code, please check again @rnetser @myakove

ocp_resources/image_stream.py
```diff
@@ -43,18 +43,21 @@ def to_dict(self):
...
-                    "tags": {
-                        "from": {"kind": "DockerImage", "name": self.image_repository},
-                        "name": self.tag,
-                        "referencePolicy": {"type": "Source"},
-                    },
-                }
+        tags = []
+        tag_name = self.tag
+        repository_name = self.image_repository
+        for x, y in zip(tag_name, repository_name):
```
myakove, line 49: I think the user should pass the tags dict.

myakove (changes requested):

ocp_resources/image_stream.py
duyanyan: Fixed.

ocp_resources/image_stream.py
```diff
@@ -14,8 +14,6 @@ def __init__(
         name=None,
         namespace=None,
         client=None,
-        image_repository=None,
-        scheduled=None,
         lookup_policy=False,
         tag=None,
```
rnetser, line 18:
```suggestion
        tags=None,
```

rnetser (changes requested):

ocp_resources/image_stream.py
duyanyan: Fixed

rnetser (approved):

duyanyan:
!verified

ocp_resources/image_stream.py
```diff
@@ -34,27 +32,14 @@ def __init__(
...
-                "spec": {
-                    "lookupPolicy": {"local": self.lookup_policy},
-                    "tags": {
-                        "from": {"kind": "DockerImage", "name": self.image_repository},
-                        "name": self.tag,
-                        "referencePolicy": {"type": "Source"},
-                    },
-                }
-            }
+            {"spec": {"lookupPolicy": {"local": self.lookup_policy}, "tags": self.tags}}
```
myakove, line 43: ```"tags": self.tags}```
Is this OK to pass tags: None? Or do we need to do:

```
if tags:
```

duyanyan: no, tag should be defined, or else the image stream doesn't work well

myakove (approved):

myakove:
@duyanyan Please rebase

duyanyan:
> @duyanyan Please rebase

Done

myakove:
!verified

myakove:
!verified

PR #467, opened 2022-03-27
Add wait for status for data source object

RoniKishner:
##### Short description:
This PR adds wait_for_status function to data source object

##### More details:
Data source object gets its status from the conditions parameters, a different approach then normal is needed.
The main wait_for_status function was copied from resource.py with only the condition parameters changed

### DIFF
--- a/ocp_resources/data_source.py
+++ b/ocp_resources/data_source.py
@@ -72,12 +72,15 @@ class DataSource(NamespacedResource):
         Wait for data source to be in status
 
         Args:
-            status: Expected status: True for a running VM, None for a stopped VM.
+            status: Expected status: True for a Ready dataSource.
             timeout (int): Time to wait for the resource.
 
         Raises:
             TimeoutExpiredError: If timeout reached.
         """
+        LOGGER.info(
+            f"Wait for {self.kind} {self.name} status to be {'' if status else 'not '}ready"
+        )
         samples = get_resource_timeout_sampler(
             resource=self, status=status, timeout=timeout, sleep=sleep
         )
--- a/ocp_resources/data_source.py
+++ b/ocp_resources/data_source.py
@@ -6,6 +6,7 @@ from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
 from ocp_resources.resource import NamespacedResource
 from ocp_resources.utils import get_resource_timeout_sampler
 
+
 LOGGER = get_logger(name=__name__)
 
 
@@ -77,9 +78,11 @@ class DataSource(NamespacedResource):
         Raises:
             TimeoutExpiredError: If timeout reached.
         """
-        samples = get_resource_timeout_sampler(self, status, timeout, sleep)
+        samples = get_resource_timeout_sampler(
+            resource=self, status=status, timeout=timeout, sleep=sleep
+        )
         for sample in samples:
             if sample.items[0].status:
                 for condition in sample.items[0].status.conditions:
-                    if condition.reason == 'Ready' and condition.status == status:
+                    if condition.reason == "Ready" and condition.status == status:
                         return
--- a/ocp_resources/data_source.py
+++ b/ocp_resources/data_source.py
@@ -1,10 +1,10 @@
 from openshift.dynamic.exceptions import ResourceNotFoundError
 
-from ocp_resources.constants import TIMEOUT_4MINUTES
+from ocp_resources.constants import TIMEOUT_4MINUTES, PROTOCOL_ERROR_EXCEPTION_DICT
 from ocp_resources.logger import get_logger
 from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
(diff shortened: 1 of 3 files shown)


ocp_resources/data_source.py
```diff
@@ -65,3 +65,31 @@ def pvc(self):
...
+            status: Expected status: True for a running VM, None for a stopped VM.
+            timeout (int): Time to wait for the resource.
+
+        Raises:
+            TimeoutExpiredError: If timeout reached.
+        """
+        LOGGER.info(
+            f"Wait for {self.kind} {self.name} status to be {'ready' if status == True else status}"
+        )
+        samples = TimeoutSampler(
```
myakove, line 83: Please extract `TimeoutSampler` code to function and use it in both places. (with the logger)

myakove (changes requested):

RoniKishner:
!verified

RoniKishner:
!verified

ocp_resources/utils.py
```diff
@@ -218,3 +219,19 @@ def remaining_time(self):
         Return the remaining part of timeout since the object was created.
         """
         return self.start_time + self.timeout - time.time()
+
+
+def get_resource_timeout_sampler(resource, status, timeout, sleep):
```
myakove, line 224: `get_resource_timeout_sampler` rename to match the action
`get_resource_status_timeout_sampler`

myakove (changes requested):

ocp_resources/utils.py
```diff
@@ -218,3 +219,19 @@ def remaining_time(self):
...
+    LOGGER.info(
+        LOGGER.info(
+            f"Wait for {resource.kind} {resource.name} status to be {'ready' if status == True else status}"
```
rnetser, line 227: 1. This is only applicable to VMs
2. Please remove the log back to where it was; this function retuns a sampler
3. There are more instances of this code in the repo; please re-use

ocp_resources/data_source.py
```diff
@@ -65,3 +66,23 @@ def pvc(self):
                 f"dataSource {self.name} is pointing to a non-existing PVC, name: {pvc_name}, "
                 f"namespace: {pvc_namespace}"
             )
+
+    def wait_for_status(self, status, timeout=TIMEOUT_4MINUTES, sleep=1):
+        """
+        Wait for data source to be in status
+
+        Args:
+            status: Expected status: True for a running VM, None for a stopped VM.
```
rnetser, line 75: This is not relevant for DataSource

ocp_resources/utils.py
RoniKishner: This function doesn't necessary gives you the status, since the status request is being done afterwards.
The Logger as well is a bit misleading since it expects you to get the status, while some 1 might get something else using this function.
It might be better to split the logger msg and the timeoutSampler creation code into 2 separate functions and use each of them on wait_for_status:
```
def get_resource_timeout_sampler
def print_status_log
```
What do you think?

ocp_resources/data_source.py
```diff
@@ -65,3 +66,23 @@ def pvc(self):
                 f"dataSource {self.name} is pointing to a non-existing PVC, name: {pvc_name}, "
                 f"namespace: {pvc_namespace}"
             )
+
+    def wait_for_status(self, status, timeout=TIMEOUT_4MINUTES, sleep=1):
```
rnetser, line 70: why can't ocp_resources.resource.Resource.wait_for_condition be used?

rnetser (changes requested):

RoniKishner:
!verified

myakove (approved):

ocp_resources/data_source.py
```diff
@@ -65,3 +66,26 @@ def pvc(self):
...
+        """
+        Wait for data source to be in status
```
rnetser, line 72:
```suggestion
        Wait for data source to be in status <xxxx>
```

ocp_resources/data_source.py
```diff
@@ -65,3 +66,26 @@ def pvc(self):
...
+
+        Args:
+            status: Expected status: True for a Ready dataSource.
+            timeout (int): Time to wait for the resource.
+
+        Raises:
+            TimeoutExpiredError: If timeout reached.
+        """
+        LOGGER.info(
+            f"Wait for {self.kind} {self.name} status to be {'' if status else 'not '}ready"
```
rnetser, line 82: Log the status (no if else)

ocp_resources/data_source.py
rnetser: you should enhance ocp_resources.resource.Resource.wait_for_condition
add optional reason and if passed, match is as well

rnetser (changes requested):

RoniKishner:
This PR can be solved by using resource.wait_for_condition(condition="Ready", status="True")

PR #482, opened 2022-04-04
Only validate_create() if self.worker_pods

Anatw:


myakove:
@Anatw please rebase

Anatw:
Will create a new PR instead

PR #499, opened 2022-04-11
add wait argument to deploy()

dbasunag:

### DIFF
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -308,17 +308,14 @@ class NodeNetworkConfigurationPolicy(Resource):
         self.ipv4_ports_backup()
         self.ipv6_ports_backup()
 
-        self.create(body=self.res)
-        if wait:
-            try:
-                self.wait_for_status_success()
-                return self
-            except Exception as exp:
-                LOGGER.error(exp)
-                super().__exit__(
-                    exception_type=None, exception_value=None, traceback=None
-                )
-                raise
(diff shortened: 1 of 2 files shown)


ocp_resources/resource.py
```diff
@@ -463,8 +463,10 @@ def _sigint_handler(self, signal_received, frame):
         self.__exit__(exception_type=None, exception_value=None, traceback=None)
         sys.exit(signal_received)
 
-    def deploy(self):
+    def deploy(self, wait=False):
```
myakove, line 466: We have more resources that define wait, please add wait there as well.

myakove (changes requested):

ocp_resources/resource.py
dbasunag: I will update those resources.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -304,18 +304,21 @@ def apply(self, resource=None):
         for _ in samples:
             return
 
-    def deploy(self):
+    def deploy(self, wait=True):
```
myakove, line 307: wait should go to self.create() and not with if wait:

myakove: wait should go to self.create() and not with if wait:

myakove (changes requested):

myakove (approved):

dbasunag:
!verified

myakove (approved):

PR #506, opened 2022-04-20
Allow manipulting timeout for vmi deletion when stopping VM

AdiZav:
Allow support for slow VMI deletions

### DIFF
--- a/ocp_resources/virtual_machine.py
+++ b/ocp_resources/virtual_machine.py
@@ -89,11 +89,13 @@ class VirtualMachine(NamespacedResource):
             self.vmi.virt_launcher_pod.wait_deleted()
             return self.vmi.wait_until_running(timeout=timeout, stop_status="dummy")
 
-    def stop(self, timeout=TIMEOUT_4MINUTES, wait=False):
+    def stop(
+        self, timeout=TIMEOUT_4MINUTES, vmi_delete_timeout=TIMEOUT_4MINUTES, wait=False
+    ):
         self.api_request(method="PUT", action="stop")
         if wait:
             self.wait_for_status(timeout=timeout, status=None)
-            return self.vmi.wait_deleted()
+            return self.vmi.wait_deleted(timeout=vmi_delete_timeout)
 
     def wait_for_status(self, status, timeout=TIMEOUT_4MINUTES, sleep=1):
         """


myakove (dismissed):

AdiZav:
!verified

ocp_resources/virtual_machine.py
```diff
@@ -89,11 +89,13 @@ def restart(self, timeout=TIMEOUT_4MINUTES, wait=False):
...
 
-    def stop(self, timeout=TIMEOUT_4MINUTES, wait=False):
+    def stop(
+        self, timeout=TIMEOUT_4MINUTES, vmi_delete_timeout=TIMEOUT_4MINUTES, wait=False
+    ):
         self.api_request(method="PUT", action="stop")
         if wait:
             self.wait_for_status(timeout=timeout, status=None)
-            return self.vmi.wait_deleted()
+            return self.vmi.wait_deleted(timeout=vmi_delete_timeout)
```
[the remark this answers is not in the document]
AdiZav: I don't know. It seems reasonable to me that a VMI will take a while to delete, and less reasonable for a VM.
We can determine it's reasonable, but I don't want to miss the time a VM reaches timeout (missing on an opportunity to determine this), because VM delete_timeout was manipulated to support VMI deletion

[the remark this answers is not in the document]
AdiZav: NVM, we can manipulate VMI delete_timeout. Went over my head :)

AdiZav:
!verified

ocp_resources/constants.py
```diff
@@ -5,3 +5,4 @@
 PROTOCOL_ERROR_EXCEPTION_DICT = {ProtocolError: []}
 NOT_FOUND_ERROR_EXCEPTION_DICT = {NotFoundError: []}
 TIMEOUT_4MINUTES = 240
+TIMEOUT_8MINUTES = 480
```
rnetser, line 8: Used only in one place, should be used directly

ocp_resources/virtual_machine_instance.py
```diff
@@ -31,7 +35,7 @@ def __init__(
         client=None,
         privileged_client=None,
         yaml_file=None,
-        delete_timeout=TIMEOUT_4MINUTES,
+        delete_timeout=TIMEOUT_8MINUTES,
```
rnetser, line 38: This now changes the default behavior for anyone using the wrapper; if this is needed for tests, it should be done there.

rnetser (changes requested):

ocp_resources/constants.py
AdiZav: ACK

ocp_resources/virtual_machine_instance.py
AdiZav: But we determined 8 mins to be a reasonable timeout for VMI deletion. Was there logic to the previous default timeout  except for 'this is what we encountered by this point'?

myakove:
@AdiZav please rebase.

ocp_resources/virtual_machine_instance.py
rnetser: it is reasonable for cnv internal usage; default should not be modified.

AdiZav: Ok. I'll revert

AdiZav:
!verified

AdiZav:
!verified

myakove:
@AdiZav please rebase

AdiZav:
!verified

myakove (approved):

rnetser (approved):

PR #527, opened 2022-05-11
add wait argument to deploy() (#499)

myakove:
* add wait argument to deploy()

* update node_network_configuration_policy.deploy()

* pass wait argument to the create call

* change default value to False


myakove:
!verified

rnetser (approved):

PR #561, opened 2022-05-24
Move kube_v1_api init to pod.execute

sarahbx:
##### Short description:
Reducing the number of attributes that must be handled
to serialize resources


sarahbx:
verified

sarahbx:
!verified

rnetser (commented):
why is this change needed? this is a breaking change.

myakove:
> why is this change needed? this is a breaking change.

it's not breaking change, but I would like to know the reason for this change.

(should be in the commit-msg)

sarahbx:
!verified

myakove (approved):

sarahbx:
!verified

myakove (approved):

sarahbx:
!verified

PR #563, opened 2022-05-25
Get api_version arg for resources

myakove:
Enable the user to set api_version when resource is initialized.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -356,6 +356,7 @@ class Resource:
         dry_run=None,
         node_selector=None,
         node_selector_labels=None,
+        api_version=None,
     ):
         """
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -356,13 +356,16 @@ def __init__(
...
         node_selector_labels=None,
+        api_version=None,
     ):
         """
         Create a API resource
 
         Args:
             name (str): Resource name
         """
+        self.api_version = api_version
```
sarahbx, line 367: if not getattr? In case api_version specified by class

sarahbx: clarification:
if api_version:
   self.api_version = ...

PR #601, opened 2022-06-01
Revert "Move kube_v1_api init to pod.execute"

myakove:
Reverts RedHatQE/openshift-python-wrapper#561

This change break pod.log() call


myakove:
!verified

PR #605, opened 2022-06-01
Move kube_v1_api to separate method

sarahbx:
Reducing the number of attributes that must be handled
to serialize resources

### DIFF
--- a/ocp_resources/mtv.py
+++ b/ocp_resources/mtv.py
@@ -60,6 +60,15 @@ class MTV:
         MIGRATION_SUCCEEDED = "The migration has SUCCEEDED."
         HOST_READY = "The host is ready."
 
+    class ConditionCategory:
+        CRITICAL = "Critical"
+
+    class ConditionType:
+        SUCCEEDED = "Succeeded"
+        FAILED = "Failed"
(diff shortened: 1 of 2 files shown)


ocp_resources/pod.py
```diff
@@ -167,3 +166,6 @@ def node(self):
     @property
     def ip(self):
         return self.instance.status.podIP
+
+    def _get_kube_v1_api(self):
```
dbasunag, line 170: This should be moved to resource.py, that way it could be reused here as well as in _collect_pod_logs()

dbasunag (changes requested):

ocp_resources/pod.py
```diff
@@ -167,3 +166,6 @@ def node(self):
     @property
     def ip(self):
         return self.instance.status.podIP
+
```
myakove, line 169: can be property

myakove (changes requested):

myakove (approved):

dbasunag (approved):

rnetser (approved):

dshchedr (approved):

rnetser:
!verified

PR #637, opened 2022-06-20
Move reuse resource and skip teardown to deploy/clean_up

myakove:
Move in order to support reuse resource/skip teardown for all calls (not only context manager calls)

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -453,6 +453,7 @@ class Resource:
 
     def __enter__(self):
         signal(SIGINT, self._sigint_handler)
+        return self.deploy()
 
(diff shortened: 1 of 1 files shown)


myakove:
!verified

myakove:
!verified

ocp_resources/resource.py
```diff
@@ -453,6 +453,17 @@ def to_dict(self):
 
     def __enter__(self):
         signal(SIGINT, self._sigint_handler)
```
sarahbx, line 455: Should signal handler be reverted upon __exit__?

myakove: No, we register it at start.

PR #639, opened 2022-06-21
Add spec.updateStrategy.registryPoll.interval to CatalogSource.

yossisegev:
Add Needed for installing standalone k8s-nmstate before upgrading from 4.10 to 4.11


myakove:
you are pushing to the wrong branch @yossisegev

PR #643, opened 2022-06-21
code-check: Add python 3.10 (#631)

myakove:
* code-check: Add python 3.10

* Update actions/setup-python to v4

* remove check on push


myakove:
!verified

PR #677, opened 2022-07-06
Move reuse resource and skip teardown to deploy/clean_up (#637)

myakove:
* Move reuse resource and skip teardown to deploy/clean_up

* Minor fix

* Add return self.deploy()


myakove:
!verified

PR #693, opened 2022-07-11
yaml_file created resource object has no name

sarahbx:
Cannot call instance until name has been set.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -868,6 +868,11 @@ class Resource:
         def _instance():
             return self.api.get(name=self.name)
 
+        if self.yaml_file and not self.name:
+            self.to_dict()
+            if not self.name:
+                raise ValueError(f"Resource name is {self.name}")
+
         return self.retry_cluster_exceptions(func=_instance)
 
     @property


ocp_resources/resource.py
```diff
@@ -868,6 +868,11 @@ def instance(self):
         def _instance():
             return self.api.get(name=self.name)
 
+        if self.yaml_file and not self.name:
```
myakove, line 871: I guess we have the same issue on namespace resource.

ocp_resources/resource.py
```diff
@@ -868,6 +868,11 @@ def instance(self):
...
+            self.to_dict()
+            if not self.name:
+                raise ValueError(f"Resource name is {self.name}")
```
rnetser, line 874: under `if not self.name`; `self.name` can be dropped from the exception message

rnetser (changes requested):

ocp_resources/resource.py
sarahbx: If name gets set to a non-null value, but is still boolean false, knowing the value may be helpful

myakove:
@sarahbx Is this PR still needed?

myakove:
@please reopen if needed.

PR #695, opened 2022-07-11
Add microsecond precision to log

sarahbx:
Log times of operations are overlapping in seconds and milliseconds

Reorder log output so datetime is at beginning of entry

Current:
```
ocp_resources.resource 2022-07-11 13:44:01 INFO Create DaemonSet utility
```

Updated:
```
2022-07-11T13:47:03.534892 ocp_resources.resource INFO Create DaemonSet utility
```

### DIFF
--- a/ocp_resources/logger.py
+++ b/ocp_resources/logger.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


sarahbx:
!verified

sarahbx:
!verified

PR #731, opened 2022-08-04
Do not require live cluster only for create class instance

myakove:
There is no need to force to have live cluster just to create any resource class instance.


rnetser (dismissed):

rnetser (approved):

myakove:
!verified

myakove:
!verified

myakove:
!cherry-pick v4.11

PR #762, opened 2022-08-11
Do not require live cluster only for create class instance (#731) (#760)

myakove:
* Do not require live cluster only for create class instance

* set version and client when needed

Co-authored-by: Meni Yakove <441263+myakove@users.noreply.github.com>


rnetser (approved):

myakove:
!verified

myakove:
!cherry-pick v4.9

PR #799, opened 2022-09-05
Add VirtualMachineExport resource

jpeimer:
##### Short description:
Add VirtualMachineExport resource

##### More details:
About VirtualMachineExport: https://kubevirt.io/user-guide/operations/export_api/#export-virtual-machine-volumes

### DIFF
--- a/ocp_resources/virtual_machine_export.py
+++ b/ocp_resources/virtual_machine_export.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+
+from ocp_resources.constants import TIMEOUT_1MINUTE
+from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
+from ocp_resources.resource import NamespacedResource
+from ocp_resources.virtual_machine import VirtualMachine
+from ocp_resources.virtual_machine_snapshot import VirtualMachineSnapshot
+
+
+class VirtualMachineExport(NamespacedResource):
+    """
+    VirtualMachineExport object.
+    """
+
+    api_group = NamespacedResource.ApiGroup.EXPORT_KUBEVIRT_IO
+
+    class SourceKind:
+        VM = VirtualMachine.kind
+        VM_SNAPSHOT = VirtualMachineSnapshot.kind
+        PVC = PersistentVolumeClaim.kind
+
+    def __init__(
+        self,
+        name=None,
+        namespace=None,
+        client=None,
+        teardown=True,
+        token_secret_ref=None,
+        source_api_group=None,
+        source_kind=None,
+        source_name=None,
+        timeout=TIMEOUT_1MINUTE,
+        delete_timeout=TIMEOUT_1MINUTE,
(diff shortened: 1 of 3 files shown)


ocp_resources/virtual_machine_export.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class VirtualMachineExport(NamespacedResource):
+    """
+    VirtualMachineExport object.
+    """
+
+    api_group = NamespacedResource.ApiGroup.EXPORT_KUBEVIRT_IO
+
+    class SourceKind:
```
myakove, line 16: Why this is needed here `class SourceKind:`?

myakove (changes requested):

ocp_resources/virtual_machine_export.py
jpeimer: Because VirtualMachineExport can accept only 3 types of source objects: 
- VM
- VM Snapshot
- PVC

myakove: Where this is used? In which cases do you use it?

myakove (changes requested):

myakove:
Please fix CI

ocp_resources/virtual_machine_export.py
jpeimer: When we create a VirtualMachineExport object, like here:

```
    with cluster_resource(VirtualMachineExport)(
        name=f"vmexport-from-snapshot",
        namespace="default",
        client=admin_client,
        token_secret_ref="virt-export-token",
        source_api_group=VirtualMachineSnapshot.api_group,
        source_kind=VirtualMachineExport.SourceKind.VM_SNAPSHOT,
        source_name="my-vmsnapshot",
        teardown=False,
    ) as vm_export_from_snapshot:
```

ocp_resources/virtual_machine_export.py
```diff
@@ -0,0 +1,79 @@
...
+        token_secret_ref=None,
+        source_api_group=None,
+        source_kind=None,
+        source_name=None,
+        timeout=TIMEOUT_4MINUTES,
+        delete_timeout=TIMEOUT_4MINUTES,
+        yaml_file=None,
+        **kwargs,
+    ):
+        if not token_secret_ref:
```
myakove, line 37: you shouldn't raise on this `token_secret_ref` if the user pass yaml_file.

myakove (changes requested):

ocp_resources/virtual_machine_export.py
jpeimer: right, changed it, thanks

myakove (approved):

rnetser (dismissed):

jpeimer:
I removed source_api_group from the check because it can be None for SourceKind PVC

jpeimer:
!verified

myakove (approved):

PR #823, opened 2022-09-19
Add an option to send keyword arguments when creating a Namespace

Anatw:

### DIFF
--- a/ocp_resources/namespace.py
+++ b/ocp_resources/namespace.py
@@ -25,9 +25,10 @@ class Namespace(Resource):
         teardown=True,
         label=None,
         yaml_file=None,
+        **kwargs
     ):
         super().__init__(
-            name=name, client=client, teardown=teardown, yaml_file=yaml_file
+            name=name, client=client, teardown=teardown, yaml_file=yaml_file, **kwargs
         )
         self.label = label
 


myakove:
Please update the title, the timeout wasn't changed. It was not exists. (you add it)

myakove:
> In several network test modules, the 240 seconds timeout was not enough.
> 
> ##### Short description:
> 
> ##### More details:
> 
> ##### What this PR does / why we need it:
> 
> ##### Which issue(s) this PR fixes:
> 
> ##### Special notes for reviewer:
> 
> ##### Bug:
> 

Please remove the reference for 'tests'.

ocp_resources/namespace.py
```diff
@@ -25,9 +25,14 @@ def __init__(
...
         yaml_file=None,
+        timeout=240,
     ):
         super().__init__(
-            name=name, client=client, teardown=teardown, yaml_file=yaml_file
+            name=name,
+            client=client,
+            teardown=teardown,
+            yaml_file=yaml_file,
+            timeout=timeout,
```
myakove, line 35: Please pass kwargs instead of timeout

ocp_resources/namespace.py
```diff
@@ -25,9 +25,14 @@ def __init__(
         teardown=True,
         label=None,
         yaml_file=None,
+        timeout=240,
```
myakove, line 28: What is the default timeout in Resource?

myakove (changes requested):

rnetser (changes requested):
This repo is not tied to openshift virtualization; please update the description

ocp_resources/namespace.py
Anatw: 240 seconds. I'll import it from Resource to make it bond to it.

ocp_resources/namespace.py
Anatw: ack.

ocp_resources/namespace.py
```diff
@@ -1,6 +1,5 @@
 from ocp_resources.logger import get_logger
-from ocp_resources.resource import Resource
-
+from ocp_resources.resource import Resource, TIMEOUT
```
dbasunag, line 2: TIMEOUT is not needed.

dbasunag (changes requested):

rnetser (changes requested):
Please update the PR's title and description; it is not related to timeout

ocp_resources/namespace.py
Anatw: I want to use it instead of '240' so that the two will be bonded.

Anatw: ack.

Anatw:
!verified

rnetser:
@Anatw please see https://github.com/RedHatQE/openshift-python-wrapper/pull/823#pullrequestreview-1113746624

PR #828, opened 2022-09-20
Fix add_rule in ClusterRole

myakove:
Add_rule we override by to_dict if the no rule was passed when initializing the resource

### DIFF
--- a/ocp_resources/cluster_role.py
+++ b/ocp_resources/cluster_role.py
@@ -60,13 +60,11 @@ class ClusterRole(Resource):
         if not self.res:
             self.res = super().to_dict()
 
-        rule = {}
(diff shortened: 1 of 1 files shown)


ocp_resources/cluster_role.py
```diff
@@ -51,9 +53,9 @@ def to_dict(self):
 
     def add_rule(
         self,
+        verbs,
```
rnetser, line 56: `if verbs:` should be removed (line 68)

rnetser (changes requested):

myakove:
!verified

rnetser (approved):

myakove:
!cherry-pick v4.12

PR #943, opened 2022-11-03
create(), remove body args

myakove:
##### Short description:
body is not needed when calling create, we use to_dict to generate the resource body.


rnetser (approved):

myakove:
!verified

myakove:
!cherry-pick v4.12

PR #971, opened 2022-11-14
rename all res to self.res

myakove:
* rename all res to self.res

* Remove return self.res

* fix self.res calls

* remove return self.res leftover

* remove return self.res leftover


rnetser (approved):

rnetser (approved):

myakove:
!verified

PR #979, opened 2022-11-17
Adding CONTRIBUTING.md file

rnetser:
##### Short description:
Add CONTRIBUTING.md with an explanation for new module (resource) addition.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,89 @@
+# Welcome to openshift-python-wrapper contributing guide
+
+Thank you for contributing to our project!  
+[Project maintainers](https://github.com/RedHatQE/openshift-python-wrapper/blob/main/OWNERS)
+
+## New contributor guide
+
+To get an overview of the project, read the [README](README.md).
+
+## Issues
+
+### Create a new issue
+If you find a problem with the code, [search if an issue already exists](https://github.com/RedHatQE/openshift-python-wrapper/issues).  
+If you open a pull request to fix the problem, an issue will ba automatically created.  
+If a related issue doesn't exist, you can open a new issue using a relevant [issue form](https://github.com/RedHatQE/openshift-python-wrapper/issues/new/choose).
+
+## Pull requests
+To contribute code to the project:
+- Fork the project and work on your forked repository
+- Before submitting a new pull request, make sure you have `pre-commit` installed  
+```bash
+pre-commit install
+```
+- When submitting a pull request, make sure to fill all the required, relevant fields for your PR.  
+Make sure the title is descriptive and short.
+- If the fix is needed in a released version, once your pull request is merged, cherry-pick it to the relevant branch(s).  
+Add `!cherry-pick <target branch to cherry-pick to>` to the PR comment.
+
+## Adding a new module (resource)
+To support working with a new (or missing) resource:
+- Add a new file under `ocp_resources`, names as the resource.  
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
+# Welcome to openshift-python-wrapper contributing guide
```
dbasunag, line 1: It looks good. Examples are to the point but helpful. 

I think it would be good to add some more information for people outside the organization/openshift-virtualization working with this repo:

1) Whats  is the process to follow when someone finds a bug? or needs to enhance a module? Should they directly open a PR or discuss with the maintainers first? Or start with logging an issue?
2) How do they reach the maintainers for questions?

These are not really a problem for us, but for someone not part of same org/starting new, it may not be easy to figure it out.

Also, if there is any requirement from the PR itself, it would be good to have it documented here. e.g. description field is added to a release, and it should be concise and clear one line summary. 

For a new resource, is there any requirement of associated docs?

rnetser: Added a section about `issues` and `pull requests`, please let me know what you think

rnetser: `For a new resource, is there any requirement of associated docs?`  see `Add a link to the resource's API reference` - do you think we need more?

CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+
+## New contributor guide
+
+To get an overview of the project, read the [README](README.md).
+
+## Issues
+
+### Create a new issue
+If you find a problem with the code, [search if an issue already exists](https://github.com/RedHatQE/openshift-python-wrapper/issues).  
+If a related issue doesn't exist, you can open a new issue using a relevant [issue form](https://github.com/RedHatQE/openshift-python-wrapper/issues/new/choose).
```
myakove, line 14: Worth to mention that if someone pushes PR then issues will be created automatically for the PR, so in this case no need to open issue.

README.md
```diff
@@ -151,9 +151,3 @@ with NodeNetworkConfigurationPolicy(
     nncp.add_interface(name="{{ capture.second-nic.interfaces.0.name }}", set_ipv4="{{ capture.first-nic.interfaces.0.ipv4 }}")
     yield nncp
 ```
-
```
myakove, line 154: I don't think we should remove this from the README

myakove (changes requested):

CONTRIBUTING.md
rnetser: done

README.md
rnetser: fine with either way; added back

CONTRIBUTING.md
myakove: `If you open a pull request to fix the problem, an issue will ba automatically created.`
should be before line 14

myakove (changes requested):

CONTRIBUTING.md
rnetser: done

rnetser:
!verified

CONTRIBUTING.md
dbasunag: Thank you for adding the sections. Both looks great! For the new resource, `Add a link to the resource's API reference` is good enough. I also think it might be good to have information on doc header styles, if anyone decides to add any. This is relevant for all codes though.

dbasunag (approved):

PR #986, opened 2022-11-29
Fix skip_existing_resource_creation_teardown to_dict call

myakove:
##### Short description:
to_dict do not return anything, store the info is self.res
Fix skip_existing_resource_creation_teardown to use it


sarahbx:
!verified

myakove:
!cherry-pick v4.12

PR #990, opened 2022-11-30
add retry on exception and timeout_seconds for get call

dbasunag:
##### Short description: 
rpc error should be retried as during some operations, they are common. However they recoverable. 
We need to add a timeout_seconds for regular get calls, as sometimes, we can have hung call.


ocp_resources/resource.py
```diff
@@ -334,6 +336,7 @@ def __init__(
         node_selector_labels=None,
         config_file=None,
         context=None,
+        timeout_seconds=60,
```
myakove, line 339: Please give a meaningful name, maybe `api_call_timeout`?

myakove (changes requested):

ocp_resources/resource.py
dbasunag: I have to keep the name, as it is expected by the api to be timeout_seconds.

dbasunag:
@rnetser when I re-requested review from @myakove , I see it removed you as reviewer. I have no idea why and how it happened. Can you please add yourself as a reviewer, as I would prefer to have both your review?

rnetser (approved):

rnetser (commented):
Please fix ci :)

dbasunag:
!verified

rnetser (approved):

ocp_resources/resource.py
myakove: Then please add docstring, we have `timeout` and `timeout_seconds` that may confuse the user.

myakove (changes requested):

ocp_resources/resource.py
dbasunag: I agree. I will add docstring.

myakove (approved):
We need to improve the docstring, but we will do it on another PR.

ocp_resources/resource.py
```diff
@@ -498,8 +515,10 @@ def _prepare_resources(cls, dyn_client, singular_name, *args, **kwargs):
 
         get_kwargs = {"singular_name": singular_name} if singular_name else {}
         return dyn_client.resources.get(
-            kind=cls.kind, api_version=cls.api_version, **get_kwargs
-        ).get(*args, **kwargs)
+            kind=cls.kind,
+            api_version=cls.api_version,
+            **get_kwargs,
+        ).get(*args, **kwargs, timeout_seconds=60)
```
rnetser, line 521:
```suggestion
        ).get(*args, **kwargs, timeout_seconds=self.timeout_seconds)
```

rnetser (changes requested):

ocp_resources/resource.py
myakove: :)

myakove (approved):

ocp_resources/resource.py
dbasunag: I took the suggestions too fast. _prepare_resources() is a class method so this won't work. and I can't use cls. timeout_seconds here.

dbasunag:
!verified

ocp_resources/resource.py
rnetser: @dbasunag `cls.` should work, you need to define it outside `__init__`

dbasunag: Ok, added it as a class attribute with a default value.

dbasunag:
!verified

rnetser (approved):

myakove (approved):

myakove:
!verified

myakove (approved):

dbasunag:
!cherry-pick v4.11

dbasunag:
cherry-pick v4.10

PR #945, opened 2022-11-03
rename all res to self.res

myakove:
Rename all res to self.res to be consistent with the code.

PR #1002, opened 2022-12-05
fix usage of self.to_dict() in some methods

vsibirsk:
##### Short description:
Fix usage of to_dict() method in several methods

PR https://github.com/RedHatQE/openshift-python-wrapper/pull/945 updated to_dict() method so it doesn't return anything. This caused issues in several methods that relied on it to have a return value

### DIFF
--- a/ocp_resources/deployment.py
+++ b/ocp_resources/deployment.py
@@ -21,7 +21,8 @@ class Deployment(NamespacedResource):
         Returns:
             Deployment is updated successfully
         """
-        body = super().to_dict()
+        super().to_dict()
+        body = self.res
         body.update({"spec": {"replicas": replica_count}})
 
         self.logger.info(f"Set deployment replicas: {replica_count}")
--- a/ocp_resources/machine_set.py
+++ b/ocp_resources/machine_set.py
@@ -177,7 +177,8 @@ class MachineSet(NamespacedResource):
         Returns:
             bool: True if scaling the machine-set was successful or wait=False, False otherwise.
         """
-        body = super().to_dict()
+        super().to_dict()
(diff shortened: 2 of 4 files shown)


rnetser (approved):

vsibirsk:
!verified

ocp_resources/deployment.py
```diff
@@ -21,7 +21,8 @@ def scale_replicas(self, replica_count=int):
         Returns:
             Deployment is updated successfully
         """
-        body = super().to_dict()
+        super().to_dict()
+        body = self.res
```
myakove, line 25: No need to save `body = self.res`, update `self.res` directly.

ocp_resources/machine_set.py
```diff
@@ -177,7 +177,8 @@ def scale_replicas(
         Returns:
             bool: True if scaling the machine-set was successful or wait=False, False otherwise.
         """
-        body = super().to_dict()
+        super().to_dict()
+        body = self.res
```
myakove, line 181: ditto

myakove (changes requested):

ocp_resources/deployment.py
vsibirsk: done

ocp_resources/machine_set.py
vsibirsk: done

vsibirsk:
!verified

myakove (approved):

myakove:
@vsibirsk Please cherry-pick to both v4.12 and v4.11

kbidarkar:
!cherry-pick v4.12

PR #1047, opened 2022-12-27
add delete after completion annotation to dv

dalia-frank:

### DIFF
--- a/ocp_resources/datavolume.py
+++ b/ocp_resources/datavolume.py
@@ -94,7 +94,6 @@ class DataVolume(NamespacedResource):
         DataVolume object
 
         Args:
-            eviction (bool, default False): If True, set evictionStrategy to LiveMigrate explicitly.
             name (str): DataVolume name.
             namespace (str): DataVolume namespace.
             source (str): source of DV - upload/http/pvc/registry.
@@ -104,14 +103,15 @@ class DataVolume(NamespacedResource):
             content_type (str, default: "kubevirt"): DataVolume content type.
             access_modes (str, default: None): DataVolume access mode.
             cert_configmap (str, default: None): name of config map for TLS certificates.
-            secret (Secret, default: None): secretRef.
+            secret (Secret, default: None): to be set as secretRef.
             client (DynamicClient): DynamicClient to use.
             volume_mode (str, default: None): DataVolume volume mode.
             hostpath_node (str, default: None): Node name to provision the DV on.
             source_pvc (str, default: None): PVC name for when cloning the DV.
             source_namespace (str, default: None): PVC namespace for when cloning the DV.
             multus_annotation (str, default: None): network nad name.
-            bind_immediate_annotation (bool, default: None): when wffc sc is used and DV should be bind immediate.
+            bind_immediate_annotation (bool, default: None): when WaitForFirstConsumer is set in  StorageClass and DV
+            should be bound immediately.
             preallocation (bool, default: None): preallocate disk space.
             teardown (bool, default: True): Indicates if this resource would need to be deleted.
             privileged_client (DynamicClient, default: None): Instance of Dynamic client
--- a/ocp_resources/datavolume.py
+++ b/ocp_resources/datavolume.py
@@ -90,6 +90,36 @@ class DataVolume(NamespacedResource):
         delete_after_completion=None,
         **kwargs,
     ):
+        """
+        DataVolume object
+
+        Args:
+            eviction (bool, default False): If True, set evictionStrategy to LiveMigrate explicitly.
+            name (str): DataVolume name.
+            namespace (str): DataVolume namespace.
+            source (str): source of DV - upload/http/pvc/registry.
+            size (str): DataVolume size - format size+size unit, for example: "5Gi".
+            storage_class (str, default: None): storage class name for DataVolume.
+            url (str, default: None): url for importing DV, when source is http/registry.
+            content_type (str, default: "kubevirt"): DataVolume content type.
+            access_modes (str, default: None): DataVolume access mode.
+            cert_configmap (str, default: None): name of config map for TLS certificates.
+            secret (Secret, default: None): secretRef.
+            client (DynamicClient): DynamicClient to use.
+            volume_mode (str, default: None): DataVolume volume mode.
+            hostpath_node (str, default: None): Node name to provision the DV on.
+            source_pvc (str, default: None): PVC name for when cloning the DV.
+            source_namespace (str, default: None): PVC namespace for when cloning the DV.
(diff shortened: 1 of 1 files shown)


myakove (changes requested):

ocp_resources/datavolume.py
```diff
@@ -87,8 +87,39 @@ def __init__(
...
+            source_namespace (str, default: None): PVC namespace for when cloning the DV.
+            multus_annotation (str, default: None): network nad name.
+            bind_immediate_annotation (bool, default: None): when wffc sc is used and DV should be bind immediate.
+            preallocation (bool, default: None): preallocate disk space.
+            teardown (bool, default: True): Indicates if this resource would need to be deleted.
+            privileged_client (DynamicClient, default: None): Instance of Dynamic client
+            yaml_file (yaml, default: None): yaml file for the resource.
+            delete_timeout (int, default: 4 minutes): timeout associated with delete action.
+            api_name (str, default: "pvc"): api used for DV, pvc/storage
+            delete_after_completion (str, default: None): annotation for garbage collector, default behavior is "true"
```
rnetser, line 121: `default behavior is "true"` -where is this configured?

ocp_resources/datavolume.py
```diff
@@ -172,6 +202,12 @@ def to_dict(self):
                 }
             if self.preallocation is not None:
                 self.res["spec"]["preallocation"] = self.preallocation
+            if self.delete_after_completion is not None:
```
rnetser, line 205:
```suggestion
            if self.delete_after_completion:
```
`delete_after_completion` is an str

ocp_resources/datavolume.py
```diff
@@ -87,8 +87,39 @@ def __init__(
...
         delete_timeout=TIMEOUT_4MINUTES,
         api_name="pvc",
+        delete_after_completion=None,
         **kwargs,
     ):
+        """
+        DataVolume object
+
+        Args:
+            eviction (bool, default False): If True, set evictionStrategy to LiveMigrate explicitly.
```
rnetser, line 97: please order the docstring by args orders

ocp_resources/datavolume.py
```diff
@@ -87,8 +87,39 @@ def __init__(
...
+            name (str): DataVolume name.
+            namespace (str): DataVolume namespace.
+            source (str): source of DV - upload/http/pvc/registry.
+            size (str): DataVolume size - format size+size unit, for example: "5Gi".
+            storage_class (str, default: None): storage class name for DataVolume.
+            url (str, default: None): url for importing DV, when source is http/registry.
+            content_type (str, default: "kubevirt"): DataVolume content type.
+            access_modes (str, default: None): DataVolume access mode.
+            cert_configmap (str, default: None): name of config map for TLS certificates.
+            secret (Secret, default: None): secretRef.
```
rnetser, line 107:
```suggestion
            secret (Secret, default: None): to be set as secretRef.
```

ocp_resources/datavolume.py
```diff
@@ -87,8 +87,39 @@ def __init__(
...
+            client (DynamicClient): DynamicClient to use.
+            volume_mode (str, default: None): DataVolume volume mode.
+            hostpath_node (str, default: None): Node name to provision the DV on.
+            source_pvc (str, default: None): PVC name for when cloning the DV.
+            source_namespace (str, default: None): PVC namespace for when cloning the DV.
+            multus_annotation (str, default: None): network nad name.
+            bind_immediate_annotation (bool, default: None): when wffc sc is used and DV should be bind immediate.
```
rnetser, line 114:
```suggestion
            bind_immediate_annotation (bool, default: None): when WaitForFirstConsumer is set in  StorageClass and DV should be bound immediately.
```

rnetser (changes requested):

ocp_resources/datavolume.py
dalia-frank: this is be mistake, fixed

ocp_resources/datavolume.py
dalia-frank: ack

ocp_resources/datavolume.py
dalia-frank: ack

ocp_resources/datavolume.py
dalia-frank: by the cdi

ocp_resources/datavolume.py
dalia-frank: all the fields that are not mandatory are set to None here, most of them are strings. should I change it?

ocp_resources/datavolume.py
rnetser: is it something an admin an change?

ocp_resources/datavolume.py
rnetser: depends on the use case; is an empty string, for example, acceptable?

ocp_resources/datavolume.py
myakove: `default behavior is "true"` I think you should remove this from the doc
we cannot update this based on CDI changes and it may not be true in later versions

ocp_resources/datavolume.py
myakove: we should not force/check the user input, the docstring should state that "true" and "false" are acceptable inputs

myakove (changes requested):

ocp_resources/datavolume.py
dalia-frank: removed

dalia-frank:
!verified

ocp_resources/datavolume.py
myakove: it is still there

myakove (changes requested):

ocp_resources/datavolume.py
dalia-frank: fixed

ocp_resources/datavolume.py
rnetser: @dalia-frank please address this comment and modify the code

rnetser (changes requested):

myakove (approved):

ocp_resources/datavolume.py
myakove: @rnetser already modified
`if self.delete_after_completion:`

myakove (approved):

dalia-frank:
!verified

rnetser (approved):

myakove (approved):

myakove:
!verified

PR #1057, opened 2023-01-10
Return from nncp_wait_for_configuration if configured successfully

Anatw:
Refactor wait_for_confinuration_conditions_unknown_or_progressing to return if nncp is already configured successfully. It can happen that when debugging, a nncp is configured successfully, but an assert raises if we get into wait_for_configuration_conditions_unknown_or_progressing after the nncp was already configured successfully.

Signed-off-by: Anat Wax <awax@redhat.com>

### DIFF
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -345,6 +345,12 @@ class NodeNetworkConfigurationPolicy(Resource):
                 return condition["reason"]
 
     def wait_for_configuration_conditions_unknown_or_progressing(self, wait_timeout=30):
+        if (
+            self.status
+            and self.status == self.Conditions.Reason.SUCCESSFULLY_CONFIGURED
+        ):
+            return
+
         samples = TimeoutSampler(
             wait_timeout=wait_timeout,
             sleep=1,


Anatw:
!verified

Anatw:
!verified

ocp_resources/node_network_configuration_policy.py
```diff
@@ -345,6 +345,12 @@ def status(self):
                 return condition["reason"]
 
     def wait_for_configuration_conditions_unknown_or_progressing(self, wait_timeout=30):
+        if (
```
myakove, line 348: This function wait for unknown_or_progressing
it doesn't make sense to return when SUCCESSFULLY_CONFIGURED

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
```diff
@@ -345,6 +345,12 @@ def status(self):
...
+            self.status
+            and self.status == self.Conditions.Reason.SUCCESSFULLY_CONFIGURED
```
rnetser, line 350: if you change nncp, this may exit before configuration changes started and may end up not waiting for the config to start

rnetser (changes requested):

myakove:
Please reopen if needed.

PR #1060, opened 2023-01-12
VirtualMachineSnapshotContent and VolumeSnapshotContent resource

jpeimer:
##### Short description:
Add VirtualMachineSnapshotContent and VolumeSnapshotContent resources

### DIFF
--- a/ocp_resources/virtual_machine_snapshot_content.py
+++ b/ocp_resources/virtual_machine_snapshot_content.py
@@ -1,0 +1,10 @@
+from ocp_resources.resource import NamespacedResource
+
+
+class VirtualMachineSnapshotContent(NamespacedResource):
+    """
+    VirtualMachineSnapshotContent object. API reference:
+    https://docs.openshift.com/container-platform/4.11/virt/virtual_machines/virtual_disks/virt-managing-vm-snapshots.html
+    """
+
+    api_group = NamespacedResource.ApiGroup.SNAPSHOT_KUBEVIRT_IO
--- a/ocp_resources/volume_snapshot_content.py
+++ b/ocp_resources/volume_snapshot_content.py
@@ -1,0 +1,10 @@
+from ocp_resources.resource import Resource
+
+
+class VolumeSnapshotContent(Resource):
+    """
+    VolumeSnapshotContent object. API reference:
+    https://docs.openshift.com/container-platform/4.11/rest_api/storage_apis/volumesnapshotcontent-snapshot-storage-k8s-io-v1.html
+    """
+
+    api_group = Resource.ApiGroup.SNAPSHOT_STORAGE_K8S_IO


rnetser (changes requested):
Please follow README on what needs to be included when adding a new resource

jpeimer:
> Please follow README on what needs to be included when adding a new resource
aah, the CONTRIBUTING, thanks, adding what's missing. 

VirtualMachineSnapshotContent - not sure if we should make an __init__ for it, from the [doc](https://docs.openshift.com/container-platform/4.11/virt/virtual_machines/virtual_disks/virt-managing-vm-snapshots.html): _It is created by the VM snapshot controller and contains references to all resources required to restore the VM._
We never create it ourselves. We want only to look at it and collect the data. 

VolumeSnapshotContent [doc](https://kubernetes.io/docs/concepts/storage/volume-snapshots/#volume-snapshot-contents):  _In dynamic provisioning, the snapshot common controller creates VolumeSnapshotContent objects._ 

So we have no intention to create any of these resources ourselves at the moment.

jpeimer:
!verified

jpeimer:
!verified

ocp_resources/virtual_machine_snapshot_content.py
```diff
@@ -0,0 +1,10 @@
+from ocp_resources.resource import NamespacedResource
+
+
+class VirtualMachineSnapshotContent(NamespacedResource):
```
myakove, line 4: Please add to_dict method with the required values to create  VirtualMachineSnapshotContent

ocp_resources/volume_snapshot_content.py
```diff
@@ -0,0 +1,10 @@
+from ocp_resources.resource import Resource
+
+
+class VolumeSnapshotContent(Resource):
```
myakove, line 4: Please add to_dict method with the required values to create  VolumeSnapshotContent

myakove (changes requested):

ocp_resources/virtual_machine_snapshot_content.py
jpeimer: VirtualMachineSnapshotContent is created by the VM snapshot controller, we don't create it ourselves.

ocp_resources/volume_snapshot_content.py
jpeimer: VolumeSnapshotContent is created by the snapshot common controller, we don't create it ourselves.

rnetser: according to https://kubernetes.io/docs/concepts/storage/volume-snapshots/#static
`A cluster administrator creates a number of VolumeSnapshotContents`
Please add to_dict

rnetser (changes requested):

ocp_resources/volume_snapshot_content.py
jpeimer: Right, it's about static provisioning, but we always prefer it to be dynamically provisioned
https://kubernetes.io/docs/concepts/storage/volume-snapshots/#volume-snapshot-contents
`In dynamic provisioning, the snapshot common controller creates VolumeSnapshotContent objects. `

rnetser: it does not matter; this is a resource that can be manually created and the code needs to support it. when using it in your implementation, you can select how to use it.

rnetser (changes requested):

myakove:
Please reopen if needed.

PR #1077, opened 2023-01-16
OCSInitialization resource from the ocs.openshift.io API group

aasserzo:

### DIFF
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -25,6 +25,9 @@ class OCSInitialization(NamespacedResource):
         enable_ceph_tools=None,
         **kwargs,
     ):
+        if not enable_ceph_tools:
+            raise ValueError("Please provide enable_ceph_tools")
+
         super().__init__(
             name=name,
             namespace=namespace,
@@ -40,10 +43,6 @@ class OCSInitialization(NamespacedResource):
     def to_dict(self):
         super().to_dict()
         if not self.yaml_file:
-            if not self.enable_ceph_tools:
-                raise ValueError(
-                    "Please provide either enable_ceph_tools or a valid YAML file"
-                )
             self.res.update(
                 {
                     "spec": {
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -1,0 +1,9 @@
+from ocp_resources.resource import NamespacedResource
+
+
+class OCSInitialization(NamespacedResource):
+    """
+    OCSInitialization object.
+    """
+
+    api_group = NamespacedResource.ApiGroup.OCS_OPENSHIFT_IO
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -9,6 +9,11 @@ class OCSInitialization(NamespacedResource):
     API reference (source code linked here due to lack of API doc for this resource):
         https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/ocsinitialization_types.go
 
+    Args:
+        enable_ceph_tools : Boolean
+            When set to True, creates a Deployment named "rook-ceph-tools" in the "openshift-storage" namespace.
+            More information about the Rook Ceph Toolbox can be found here:
+            https://rook.io/docs/rook/v1.10/Troubleshooting/ceph-toolbox/
     """
 
     api_group = NamespacedResource.ApiGroup.OCS_OPENSHIFT_IO
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -43,9 +43,7 @@ class OCSInitialization(NamespacedResource):
 
     def to_dict(self):
         super().to_dict()
-        if not self.yaml_file:
-            if not self.enable_ceph_tools:
-                raise ValueError("Please provide enable_ceph_tools")
+        if self.yaml_file is None and self.enable_ceph_tools is not None:
             self.res.update(
                 {
                     "spec": {
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -3,12 +3,6 @@ from ocp_resources.resource import NamespacedResource
 
 
 class OCSInitialization(NamespacedResource):
-    """
-    OCSInitialization object.
-
-    API reference (source code linked here due to lack of API doc for this resource):
-        https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/ocsinitialization_types.go
-    """
 
     api_group = NamespacedResource.ApiGroup.OCS_OPENSHIFT_IO
 
@@ -25,8 +19,14 @@ class OCSInitialization(NamespacedResource):
         **kwargs,
     ):
         """
-        enable_ceph_tools : Boolean
-            When set to True, creates a Deployment named "rook-ceph-tools" in the "openshift-storage" namespace.
+        OCSInitialization object.
+
+        API reference (source code linked here due to lack of API doc for this resource):
+            https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/ocsinitialization_types.go
+
+        Args:
+        enable_ceph_tools (Boolean):
+            When True, creates a Deployment named "rook-ceph-tools" in the "openshift-storage" namespace.
             More information about the Rook Ceph Toolbox can be found here:
             https://rook.io/docs/rook/v1.10/Troubleshooting/ceph-toolbox/
         """
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -30,8 +30,6 @@ class OCSInitialization(NamespacedResource):
         enable_ceph_tools=None,
         **kwargs,
     ):
-        if not enable_ceph_tools:
-            raise ValueError("Please provide enable_ceph_tools")
 
         super().__init__(
             name=name,
@@ -48,6 +46,8 @@ class OCSInitialization(NamespacedResource):
     def to_dict(self):
         super().to_dict()
         if not self.yaml_file:
+            if not self.enable_ceph_tools:
+                raise ValueError("Please provide enable_ceph_tools")
             self.res.update(
                 {
                     "spec": {
--- a/ocp_resources/ocs_initialization.py
+++ b/ocp_resources/ocs_initialization.py
@@ -8,12 +8,6 @@ class OCSInitialization(NamespacedResource):
 
     API reference (source code linked here due to lack of API doc for this resource):
         https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/ocsinitialization_types.go
-
-    Args:
-        enable_ceph_tools : Boolean
-            When set to True, creates a Deployment named "rook-ceph-tools" in the "openshift-storage" namespace.
-            More information about the Rook Ceph Toolbox can be found here:
-            https://rook.io/docs/rook/v1.10/Troubleshooting/ceph-toolbox/
     """
 
     api_group = NamespacedResource.ApiGroup.OCS_OPENSHIFT_IO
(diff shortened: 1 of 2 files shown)


ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,9 @@
+from ocp_resources.resource import NamespacedResource
```
myakove, line 1: Please follow https://github.com/RedHatQE/openshift-python-wrapper/blob/main/CONTRIBUTING.md for adding a new resource.

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: I have. Is there something specifically wrong?

myakove: https://github.com/RedHatQE/openshift-python-wrapper/blob/main/CONTRIBUTING.md#adding-a-new-module-resource

aasserzo: If you are referring to the missing __init__ function, this resource is not to be created using the API, it's only being added for patching, as stated in the commit message.
I've seen this done in many other resources in the wrapper.
Let me know if it needs to be added regardless.

myakove: It is needed, we do not add empty resources anymore.

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,35 @@
+from ocp_resources.constants import TIMEOUT_4MINUTES
```
myakove, line 1: Missing a link to the resource doc

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,35 @@
...
+            client=client,
+            teardown=teardown,
+            privileged_client=privileged_client,
+            yaml_file=yaml_file,
+            delete_timeout=delete_timeout,
+            **kwargs,
+        )
+
+    def to_dict(self):
+        super().to_dict()
```
myakove, line 35: Should have all required fields to create a resource.

ocp_resources/resource.py
```diff
@@ -299,6 +299,7 @@ class ApiGroup:
         V2V_KUBEVIRT_IO = "v2v.kubevirt.io"
         VELERO_IO = "velero.io"
         VM_KUBEVIRT_IO = "vm.kubevirt.io"
+        OCS_OPENSHIFT_IO = "ocs.openshift.io"
```
myakove, line 302: Move to the right place, ordered by alphabetic

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: There is no entry for this resource in the API doc

ocp_resources/ocs_initialization.py
aasserzo: done

ocp_resources/resource.py
aasserzo: done

ocp_resources/ocs_initialization.py
aasserzo: as discussed, added a link to the source code

ocp_resources/ocs_initialization.py
aasserzo: added as discussed

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(
+        self,
+        name=None,
+        namespace=None,
+        client=None,
+        teardown=False,
+        privileged_client=None,
+        yaml_file=None,
+        delete_timeout=TIMEOUT_4MINUTES,
+        enable_ceph_tools=False,
```
myakove, line 25: Is False is the default?
If not, please set it to None and raise if the user didn't send it.

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: Done

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,52 @@
...
+        name=None,
+        namespace=None,
+        client=None,
+        teardown=False,
+        privileged_client=None,
+        yaml_file=None,
+        delete_timeout=TIMEOUT_4MINUTES,
+        enable_ceph_tools=None,
+        **kwargs,
+    ):
```
rnetser, line 27: please add docstring (especially as `enable_ceph_tools` does not appear in the provided link)

rnetser (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: Done

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,57 @@
...
+        if not enable_ceph_tools:
```
myakove, line 33: Please move it inside `to_dict()` after `if not self.yaml_file:`

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: Done

aasserzo:
!verified

myakove (approved):

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class OCSInitialization(NamespacedResource):
+    """
+    OCSInitialization object.
+
+    API reference (source code linked here due to lack of API doc for this resource):
+        https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/ocsinitialization_types.go
+
+    Args:
```
myakove, line 12: I missed it, The docstring (`Args` part) should be in `__init__`, please fix.

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: done

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,57 @@
...
+        teardown=False,
+        privileged_client=None,
+        yaml_file=None,
+        delete_timeout=TIMEOUT_4MINUTES,
+        enable_ceph_tools=None,
+        **kwargs,
+    ):
+        """
+        enable_ceph_tools : Boolean
+            When set to True, creates a Deployment named "rook-ceph-tools" in the "openshift-storage" namespace.
```
myakove, line 29: `in the "openshift-storage" namespace.`

I see that it can be on any namespace
See https://github.com/rook/rook/blob/master/deploy/examples/toolbox.yaml#L5
for example.

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,57 @@
...
+        client=None,
+        teardown=False,
+        privileged_client=None,
+        yaml_file=None,
+        delete_timeout=TIMEOUT_4MINUTES,
+        enable_ceph_tools=None,
+        **kwargs,
+    ):
+        """
+        enable_ceph_tools : Boolean
```
myakove, line 28: Docstring format is:

"""
Description

Args:
    arg1 (type): description

""" 

I think the best is to move all docstring from the class to the init, then you will have description and you will add the Args

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: It can, but by flipping this boolean, it's created in 'openshift-storage'. Similarly, the name can be different as well, but in this case it's "rook-ceph-tools".

ocp_resources/ocs_initialization.py
aasserzo: done

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,57 @@
...
+            yaml_file=yaml_file,
+            delete_timeout=delete_timeout,
+            **kwargs,
+        )
+        self.enable_ceph_tools = enable_ceph_tools
+
+    def to_dict(self):
+        super().to_dict()
+        if not self.yaml_file:
+            if not self.enable_ceph_tools:
```
rnetser, line 49:
```suggestion
            if self.enable_ceph_tools is not None:
```
otherwise it will also fail if `enable_ceph_tools` is False

rnetser (changes requested):
see comment in code

ocp_resources/ocs_initialization.py
myakove: OK

ocp_resources/ocs_initialization.py
aasserzo: Now that I come to think of it, this parameter is not required for the resource to be successfully created/applied. I will remove this condition.

aasserzo:
!verified

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,53 @@
...
+            privileged_client=privileged_client,
+            yaml_file=yaml_file,
+            delete_timeout=delete_timeout,
+            **kwargs,
+        )
+        self.enable_ceph_tools = enable_ceph_tools
+
+    def to_dict(self):
+        super().to_dict()
+        if self.yaml_file is None and self.enable_ceph_tools is not None:
```
myakove, line 46: `if self.yaml_file is None` Please keep our code convention
`if not self.yaml_file`

myakove (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: done

aasserzo:
!verified

ocp_resources/ocs_initialization.py
```diff
@@ -0,0 +1,53 @@
...
+            privileged_client=privileged_client,
+            yaml_file=yaml_file,
+            delete_timeout=delete_timeout,
+            **kwargs,
+        )
+        self.enable_ceph_tools = enable_ceph_tools
+
+    def to_dict(self):
+        super().to_dict()
+        if not self.yaml_file and self.enable_ceph_tools is not None:
```
rnetser, line 46: i may have been confused before but this is what we need right?
```suggestion
        if not self.yaml_file and self.enable_ceph_tools:
```

?

aasserzo: Your comment was correct, but the suggestion contained wrong code :smile: 
Your current suggestion would mean that the condition would be True only if enable_ceph_tools is true, which is what bothered you in the first place.
'not None' is the correct logic here, because I want the condition to be met when enable_ceph_tools exists, regardless of if it's True or False.

rnetser: so what confused me was the docstring:
```
enable_ceph_tools (bool): When True, creates a Deployment named "rook-ceph-tools" in the "openshift-storage"  namespace.
```
which should be updated based on your last comment

rnetser (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: The OCSInitialization resource always exists. I'm adding it to the wrapper so that I may apply the "enableCephTools = True" spec.
When "enableCephTools = True" is added to the spec, a new Deployment is created (aside from this OCSInitialization resource).

So to me the docstring makes sense, as well as the init function. What update would you like to see?

rnetser: If the use case is to support "enableCephTools = True" then the condition should be  `if not self.yaml_file and self.enable_ceph_tools` because there's no point in adding `"enableCephTools = False" `

rnetser (changes requested):

ocp_resources/ocs_initialization.py
aasserzo: There might not be a point, but it is supported by the API.
Also, this is a boolean parameter, this both False and True should be supported, even if deleting the parameter altogether doesn't differ functionally from setting it to False.

rnetser: when setting "enableCephTools = False"; will it be present in the CR spec or will it be removed?

rnetser: if present as False then LGTM

rnetser (approved):

myakove (approved):

PR #1058, opened 2023-01-11
add ForbiddenError to the retry list and add default exceptions to a few calls

dbasunag:
##### More details:
add DEFAULT_CLUSTER_RETRY_EXCEPTIONS to the following calls:

1. wait
2. wait_for_status
3. wait_for_ready_status

PR #1086, opened 2023-01-18
[v4.10] add ForbiddenError to the retry list and add default exceptions to a few calls

dbasunag:
(#1058)

* add ForbiddenError to the retry list

* add default exceptions to a few calls


rnetser (approved):

dbasunag:
!verified

PR #1097, opened 2023-02-06
[v4.11] Fix skip_existing_resource_creation_teardown to_dict call

sarahbx:
##### Short description:

to_dict do not return anything, store the info is self.res
Fix skip_existing_resource_creation_teardown to use it

Picked from: https://github.com/RedHatQE/openshift-python-wrapper/pull/986


sarahbx:
!verified

myakove (approved):

PR #1116, opened 2023-02-17
Convert setup.py/cfg to pyproject.toml

myakove:

PR #1124, opened 2023-02-21
Convert setup.py/cfg to pyproject.toml (#1116) (#1122)

myakove:
* Convert setup.py/cfg to pyproject.toml

* Add VERSION file

* Set python version to >=3.7


myakove:
!verified

PR #1189, opened 2023-04-17
Add dns resource

Anatw:

### DIFF
--- a/ocp_resources/dns.py
+++ b/ocp_resources/dns.py
@@ -1,0 +1,12 @@
+# API reference:
+#       https://docs.openshift.com/container-platform/4.12/rest_api/config_apis/dns-config-openshift-io-v1.html
+
+from ocp_resources.resource import Resource
+
+
+class DNS(Resource):
+    """
+    DNS object.
+    """
+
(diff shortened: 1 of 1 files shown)


Anatw:
!verified

ocp_resources/dns.py
```diff
@@ -0,0 +1,9 @@
+from ocp_resources.resource import Resource
+
+
+class DNS(Resource):
```
jpeimer, line 4: I believe we need to follow the contributing requirements:
https://github.com/RedHatQE/openshift-python-wrapper/blob/main/CONTRIBUTING.md#adding-a-new-module-resource

jpeimer (changes requested):

ocp_resources/dns.py
rnetser: iiuc this is a resource added by k8s (a user cannot add it).
please add a link to the api

rnetser (changes requested):

ocp_resources/dns.py
Anatw: Ack.

myakove (approved):

PR #1192, opened 2023-04-18
Add label parameter to Resource object

RoniKishner:
##### Short description:
Add label parameter to Resource class

##### More details:
- Added label parameter to Resource object so it will be added to the resource manifest under metadata->labels
- Changed namespace.py so it will use the new label parameter in Resource object

Label field is relevant to all resources, this will enable easier use of creating a resource with labels.
Also made changes to Namespace so they will use the label parameter 

Enable use of label when creating a new DataSource

### DIFF
--- a/ocp_resources/namespace.py
+++ b/ocp_resources/namespace.py
@@ -17,7 +17,6 @@ class Namespace(Resource):
         name=None,
         client=None,
         teardown=True,
-        label=None,
         yaml_file=None,
         delete_timeout=TIMEOUT_4MINUTES,
         **kwargs,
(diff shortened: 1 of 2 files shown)


ocp_resources/data_source.py
```diff
@@ -17,6 +17,7 @@ def __init__(
         teardown=True,
         yaml_file=None,
         delete_timeout=TIMEOUT_4MINUTES,
+        label=None,
```
rnetser, line 20: either you add it to all resources or do not add it as we already have `kwargs` which are passed.

rnetser (changes requested):

rnetser (approved):

rnetser (changes requested):
Please update the title and description to match the updated code

rnetser (approved):

myakove (approved):

PR #1222, opened 2023-05-11
Hash sensitive data in resource logs

omrirh:
- Hash resource sensitive data in CRUD methods log messages

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -706,11 +706,10 @@ class Resource:
         if not self.res:
             self.to_dict()
 
+        hashed_res = self.hash_resource_dict(resource_dict=self.res)
         self.logger.info(f"Create {self.kind} {self.name}")
-        self.logger.info(f"Posting {self.hash_resource_dict(resource_dict=self.res)}")
-        self.logger.debug(
-            f"\n{yaml.dump(self.hash_resource_dict(resource_dict=self.res))}"
-        )
+        self.logger.info(f"Posting {hashed_res}")
+        self.logger.debug(f"\n{yaml.dump(hashed_res)}")
         resource_ = self.api.create(
             body=self.res, namespace=self.namespace, dry_run=self.dry_run
         )
@@ -725,11 +724,10 @@ class Resource:
     def delete(self, wait=False, timeout=TIMEOUT_4MINUTES, body=None):
         self.logger.info(f"Delete {self.kind} {self.name}")
         if self.exists:
-            data = self.instance.to_dict()
-            self.logger.info(f"Deleting {self.hash_resource_dict(resource_dict=data)}")
-            self.logger.debug(
-                f"\n{yaml.dump(self.hash_resource_dict(resource_dict=data))}"
-            )
+            hashed_data = self.hash_resource_dict(resource_dict=self.instance.to_dict())
+
+            self.logger.info(f"Deleting {hashed_data}")
+            self.logger.debug(f"\n{yaml.dump(hashed_data)}")
 
         try:
             res = self.api.delete(name=self.name, namespace=self.namespace, body=body)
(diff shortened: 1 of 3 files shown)


myakove (changes requested):

ocp_resources/resource.py
```diff
@@ -722,8 +723,9 @@ def delete(self, wait=False, timeout=TIMEOUT_4MINUTES, body=None):
         self.logger.info(f"Delete {self.kind} {self.name}")
         if self.exists:
             data = self.instance.to_dict()
-            self.logger.info(f"Deleting {data}")
-            self.logger.debug(f"\n{yaml.dump(data)}")
+            if self.kind != "Secret":
```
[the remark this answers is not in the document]
omrirh: done

ocp_resources/resource.py
```diff
@@ -1017,6 +1030,26 @@ def to_yaml(self):
...
+        for key in nested_dict.keys():
+            if key in self.keys_to_hash:
+                if isinstance(nested_dict[key], dict):
+                    nested_dict[key] = {_key: "***" for _key in nested_dict[key].keys()}
+                else:
+                    nested_dict[key] = "***"
+            else:
+                nested_dict[key] = self.hash_resource_dict_rec(nested_dict[key])
+
+        return nested_dict
```
rnetser, line 1051: should be saved in `self.hashed_resource_dict` and not called every time this is needed

ocp_resources/resource.py
```diff
@@ -1017,6 +1030,26 @@ def to_yaml(self):
         self.logger.info(f"\n{resource_yaml}")
         return resource_yaml
 
+    def hash_resource_dict(self, resource_dict: dict):
+        if hasattr(self, "keys_to_hash"):
+            return self.hash_resource_dict_rec(nested_dict=copy.deepcopy(resource_dict))
+        return resource_dict
+
+    def hash_resource_dict_rec(self, nested_dict: dict):
```
rnetser, line 1038: should be in one function

ocp_resources/resource.py
```diff
@@ -1017,6 +1030,26 @@ def to_yaml(self):
...
+
+        for key in nested_dict.keys():
+            if key in self.keys_to_hash:
+                if isinstance(nested_dict[key], dict):
+                    nested_dict[key] = {_key: "***" for _key in nested_dict[key].keys()}
+                else:
+                    nested_dict[key] = "***"
+            else:
+                nested_dict[key] = self.hash_resource_dict_rec(nested_dict[key])
+
```
rnetser, line 1050: no need for this; keys_to_hash should be a list of absolute paths, e.g ["spec.data", "metadata.name"]
Then use python-benedict to get the value from the dict and hash it; no need for recursive function.

ocp_resources/secret.py
```diff
@@ -73,3 +73,10 @@ def certificate_not_before(self):
         return self.instance.metadata.annotations[
             "auth.openshift.io/certificate-not-before"
         ]
+
+    @property
+    def keys_to_hash(self):
+        """
+        Resource attributes to hash in the logs
```
rnetser, line 80:
```suggestion
        Resource attributes list to hash in the logs.
        
         The list should hold absolute key paths in resource dict.
         
         Example:
             ["spec.data"]
```

ocp_resources/secret.py
```diff
@@ -73,3 +73,10 @@ def certificate_not_before(self):
...
+        """
+        return ["data"]
```
rnetser, line 82:
```suggestion
        return ["spec.data"]
```

rnetser (changes requested):

ocp_resources/resource.py
omrirh: benedict does not support dictionary keys that contain `"."` character:
```
ValueError: Key should not contain keypath separator '.', found: '.dockerconfigjson'.
```

ocp_resources/resource.py
omrirh: done

ocp_resources/resource.py
omrirh: done

ocp_resources/secret.py
omrirh: done

ocp_resources/resource.py
omrirh: currently for each CRUD method logs we pass a different data dict: Resource `delete` logs `self.instance.to_dict()`, `update` logs a given `resource_dict` and `create logs `self.res`.
Should we use a single attribute of resource as `self.hashed_resource_dict` (e.g. hash `self.instance.to_dict()) ?

myakove (dismissed):

myakove (approved):

rnetser (approved):

PR #1229, opened 2023-05-28
Expand contents of SecurityContextConstraints

yossisegev:
Support extra parameters of SecurityContextConstraints resource.

### DIFF
--- a/ocp_resources/security_context_constraints.py
+++ b/ocp_resources/security_context_constraints.py
@@ -5,32 +5,32 @@ class SecurityContextConstraints(Resource):
     api_group = Resource.ApiGroup.SECURITY_OPENSHIFT_IO
 
     def __init__(
-            self,
-            allow_host_dir_volume_plugin=False,
-            allow_host_ipc=False,
-            allow_host_network=False,
-            allow_host_pid=False,
-            allow_host_ports=False,
-            allow_privilege_escalation=None,
-            allow_privileged_container=False,
-            allowed_capabilities=None,
-            allowed_flex_volumes=None,
-            allowed_unsafe_sysctls=None,
-            default_add_capabilities=None,
-            default_allow_privilege_escalation=None,
-            forbidden_sysctls=None,
-            fs_group=None,
-            groups=None,
-            priority=0,
-            read_only_root_filesystem=False,
-            required_drop_capabilities=None,
-            run_as_user=None,
-            se_linux_context=None,
-            seccomp_profiles=None,
-            supplemental_groups=None,
-            users=None,
-            volumes=None,
-            **kwargs,
+        self,
+        allow_host_dir_volume_plugin=False,
+        allow_host_ipc=False,
+        allow_host_network=False,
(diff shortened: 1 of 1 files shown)


ocp_resources/security_context_constraints.py
```diff
@@ -3,7 +3,196 @@
 
 class SecurityContextConstraints(Resource):
     """
-    Security Context Constraints object.
+    Security Context Constraints object. API reference:
+    https://docs.openshift.com/container-platform/4.13/rest_api/security_apis/securitycontextconstraints-security-openshift-io-v1.html
+
+    Args:
```
rnetser, line 9: 1. should be under __init__
2. do not repeat the api docs
3. defaults by the api should not appear here

ocp_resources/security_context_constraints.py
```diff
@@ -3,7 +3,196 @@
...
+        self.seccomp_profiles = seccomp_profiles
+        self.supplemental_groups = supplemental_groups
+        self.users = users
+        self.volumes = volumes
+
+    def to_dict(self):
+        super().to_dict()
+        if not self.yaml_file:
+            self.res.update()
+            manifest_spec = self.res
```
rnetser, line 117: no need to save

rnetser (changes requested):

ocp_resources/security_context_constraints.py
yossisegev: Done

ocp_resources/security_context_constraints.py
yossisegev: Done

myakove (approved):

ocp_resources/security_context_constraints.py
```diff
@@ -2,8 +2,174 @@
...
+        users=None,
+        volumes=None,
+        **kwargs,
+    ):
+        """
+        Security Context Constraints object. API reference:
+        https://docs.openshift.com/container-platform/4.13/rest_api/security_apis/securitycontextconstraints-security-openshift-io-v1.html
+
+        Args:
+            allow_host_dir_volume_plugin (bool): Determines if the policy allow containers to use the HostDir volume
```
rnetser, line 40: please add defaults, e.g `allow_host_dir_volume_plugin (bool, default: False)`

ocp_resources/security_context_constraints.py
```diff
@@ -2,8 +2,174 @@
...
+            if self.forbidden_sysctls:
+                self.res["forbiddenSysctls"] = self.forbidden_sysctls
+
+            if self.fs_group:
+                self.res["fsGroup"] = self.fs_group
+
+            if self.groups:
+                self.res["groups"] = self.groups
+
+            if self.priority:
```
rnetser, line 150: `priority` default is `0`
is `0` a valid option?if so, the code below will not be reached

rnetser (changes requested):

PR #1243, opened 2023-06-15
Refactor the code according to flake8 errors (#1205)

Anatw:


rnetser (changes requested):
should be done in `main` and cherryppicked

rnetser (approved):

PR #1366, opened 2023-08-23
fix wait_until_running() call if vmi object doesn't exist

vsibirsk:
##### Short description:
fix error when calling wait_until_running() method while vmi still doesn't exist

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -708,6 +708,10 @@ class Resource:
                 self.logger.error(
                     f"Status of {self.kind} {self.name} is {current_status}"
                 )
+            else:
+                self.logger.error(
+                    f"{self.kind} {self.name}: status is None!/does not exist!"
+                )
             raise
 
     def create(self, wait=False):
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -708,10 +708,6 @@ class Resource:
                 self.logger.error(
                     f"Status of {self.kind} {self.name} is {current_status}"
                 )
-            else:
-                self.logger.error(
-                    f"{self.kind} {self.name}: status is None!/does not exist!"
-                )
             raise
 
     def create(self, wait=False):
--- a/ocp_resources/virtual_machine_instance.py
+++ b/ocp_resources/virtual_machine_instance.py
@@ -127,7 +127,8 @@ class VirtualMachineInstance(NamespacedResource):
         except TimeoutExpiredError as sampler_ex:
             if not logs:
                 raise
-            self.logger.error(f"VMI {self.name} status: {self.instance.status.phase}")
+            if not self.exists:
+                self.logger.error(f"VMI {self.name} does not exist!")
             try:
                 virt_pod = self.virt_launcher_pod
                 self.logger.error(
--- a/ocp_resources/virtual_machine_instance.py
+++ b/ocp_resources/virtual_machine_instance.py
@@ -117,10 +117,6 @@ class VirtualMachineInstance(NamespacedResource):
             TimeoutExpiredError: If VMI failed to run.
         """
         try:
-            self.wait(timeout=timeout)
-            self.logger.info(
-                f"VMI {self.name} status before wait: {self.instance.status.phase}"
-            )
             self.wait_for_status(
                 status=self.Status.RUNNING, timeout=timeout, stop_status=stop_status
             )
--- a/ocp_resources/virtual_machine_instance.py
+++ b/ocp_resources/virtual_machine_instance.py
@@ -117,10 +117,10 @@ class VirtualMachineInstance(NamespacedResource):
             TimeoutExpiredError: If VMI failed to run.
         """
         try:
-            if self.exists:
-                self.logger.info(
-                    f"VMI {self.name} status before wait: {self.instance.status.phase}"
-                )
+            self.wait(timeout=timeout)
+            self.logger.info(
+                f"VMI {self.name} status before wait: {self.instance.status.phase}"
+            )
             self.wait_for_status(
                 status=self.Status.RUNNING, timeout=timeout, stop_status=stop_status
             )
--- a/ocp_resources/virtual_machine_instance.py
+++ b/ocp_resources/virtual_machine_instance.py
@@ -123,8 +123,6 @@ class VirtualMachineInstance(NamespacedResource):
         except TimeoutExpiredError as sampler_ex:
             if not logs:
                 raise
-            if not self.exists:
-                self.logger.error(f"VMI {self.name} does not exist!")
             try:
                 virt_pod = self.virt_launcher_pod
                 self.logger.error(
--- a/ocp_resources/virtual_machine_instance.py
+++ b/ocp_resources/virtual_machine_instance.py
@@ -117,9 +117,10 @@ class VirtualMachineInstance(NamespacedResource):
             TimeoutExpiredError: If VMI failed to run.
         """
         try:
-            self.logger.info(
-                f"VMI {self.name} status before wait: {self.instance.status.phase}"
-            )
+            if self.exists:
+                self.logger.info(
+                    f"VMI {self.name} status before wait: {self.instance.status.phase}"
+                )
             self.wait_for_status(
                 status=self.Status.RUNNING, timeout=timeout, stop_status=stop_status
             )


ocp_resources/virtual_machine_instance.py
```diff
@@ -115,16 +115,20 @@ def wait_until_running(self, timeout=TIMEOUT_4MINUTES, logs=True, stop_status=No
...
                 status=self.Status.RUNNING, timeout=timeout, stop_status=stop_status
             )
         except TimeoutExpiredError:
             if not logs:
                 raise
-            self.logger.error(f"VMI {self.name} status: {self.instance.status.phase}")
+            if self.exists:
+                self.logger.error(
+                    f"VMI {self.name} status: {self.instance.status.phase}"
+                )
```
jpeimer, line 131: Should we make an else and log that vmi doesn't exist?

myakove: The user should start the vm (if it was started is should be existed) so no, no need.

Why `if self.exists:` needed here?

jpeimer: If the user started the VM, and VMI wasn't created - it's a bug and it would be nice to log this behavior. 
But @vsibirsk already checked that in this case, we'll not get to this line, and will fail on a `NotFoundError` in `wait_for_status`

vsibirsk: calling `self.instance` or `self.virt_launcher_pod` raise kubernetis api exception (404 not found in case VMI doesn't exist)
updated the except block here so that it handles this correctly and raises `TimeoutExpiredError`

myakove: Then `ocp_resources/virtual_machine.py:105` needs to be modified with an option to check if the vmi exists

vsibirsk: `VirtualMachine` is not involved here
here `self.wait_for_status`is called, not self.wait_for_ready_status (`VirtualMachineInstance` inherits from `NamespacedResource`, not from `VIrtualMachine`)
modifying `Resource.wait_for_status()` is not needed since it's not what is causing the issue

rnetser: > VirtualMachine is not involved here

It does (it is not just addressing the issue) - the right flow is to start a VM, wait for it to be ready, and **optionally** check `vmi.exists` and continue the flow.

rnetser (dismissed):

ocp_resources/virtual_machine_instance.py
vsibirsk: correct, but all of this is done outside of `VIrtualMachineInstance` methods (and VM won't ever be ready, if VMI is not ready)
this PR is needed just to handle the **unexpected** exception
i don't thinks it is a right way to call a class method and get a big traceback if something goes wrong

vsibirsk: it makes sense for `try` block. but in `except` block there's no point because we already hit timeout when vmi is not created

ocp_resources/virtual_machine_instance.py
```diff
@@ -115,17 +115,26 @@ def wait_until_running(self, timeout=TIMEOUT_4MINUTES, logs=True, stop_status=No
             TimeoutExpiredError: If VMI failed to run.
         """
         try:
-            self.logger.info(
-                f"VMI {self.name} status before wait: {self.instance.status.phase}"
-            )
+            if self.exists:
+                self.logger.info(
+                    f"VMI {self.name} status before wait: {self.instance.status.phase}"
+                )
```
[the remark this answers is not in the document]
vsibirsk: removed redundant log

ocp_resources/virtual_machine_instance.py
```diff
@@ -117,16 +117,14 @@ def wait_until_running(self, timeout=TIMEOUT_4MINUTES, logs=True, stop_status=No
...
-                f"VMI {self.name} status before wait: {self.instance.status.phase}"
-            )
             self.wait_for_status(
                 status=self.Status.RUNNING, timeout=timeout, stop_status=stop_status
             )
         except TimeoutExpiredError as sampler_ex:
             if not logs:
                 raise
-            self.logger.error(f"VMI {self.name} status: {self.instance.status.phase}")
+            if not self.exists:
```
dbasunag, line 126: Is this still needed?

ocp_resources/virtual_machine_instance.py
```diff
@@ -117,16 +117,14 @@ def wait_until_running(self, timeout=TIMEOUT_4MINUTES, logs=True, stop_status=No
...
+                self.logger.error(f"VMI {self.name} does not exist!")
```
jpeimer, line 127: I think we can remove this now, with the new changes, we'll get the `ResourceNotFoundError` if `virt_launcher_pod` doesn't exist

vsibirsk: moved this to `wait_for_status()` (so if no `current_status`, it'll show proper log)
```
2023-08-31T15:03:01.402834 ocp_resources.resource INFO kind: VirtualMachineInstance api version: kubevirt.io/v1
2023-08-31T15:03:01.402999 ocp_resources VirtualMachineInstance INFO Wait for VirtualMachineInstance container-disk-vm-1693483207-3708382 status to be Running
2023-08-31T15:03:01.403144 ocp_resources.utils INFO Waiting for 2 seconds, retry every 1 seconds. (Function: kubernetes.dynamic.client.get Kwargs: {'field_selector': 'metadata.name==container-disk-vm-1693483207-3708382', 'namespace': 'virt-general-test-container-disk-vm'})
2023-08-31T15:03:03.483983 ocp_resources VirtualMachineInstance ERROR VirtualMachineInstance container-disk-vm-1693483207-3708382: status is None!/does not exist!
*** kubernetes.dynamic.exceptions.NotFoundError: 404
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Audit-Id': '31546098-5020-48dc-8a9b-ce54caa78728', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload', 'X-Kubernetes-Pf-Flowschema-Uid': '7a14a742-a4ff-4502-8dc6-69b0943f1792', 'X-Kubernetes-Pf-Prioritylevel-Uid': '8cec7eb0-0f89-4eee-80af-f4d87bad54bd', 'Date': 'Thu, 31 Aug 2023 12:03:03 GMT', 'Content-Length': '312'})
HTTP response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"virtualmachineinstances.kubevirt.io \\"container-disk-vm-1693483207-3708382\\" not found","reason":"NotFound","details":{"name":"container-disk-vm-1693483207-3708382","group":"kubevirt.io","kind":"virtualmachineinstances"},"code":404}\n'
Original traceback: 
...
```

ocp_resources/resource.py
```diff
@@ -708,6 +708,10 @@ def wait_for_status(
                 self.logger.error(
                     f"Status of {self.kind} {self.name} is {current_status}"
                 )
+            else:
```
dbasunag, line 711: This is not useful. if there is no current status, you already know that the status does not exist.

vsibirsk: removed

dbasunag (approved):

jpeimer (approved):

ocp_resources/virtual_machine_instance.py
```diff
@@ -117,16 +117,12 @@ def wait_until_running(self, timeout=TIMEOUT_4MINUTES, logs=True, stop_status=No
             TimeoutExpiredError: If VMI failed to run.
         """
         try:
-            self.logger.info(
```
myakove, line 120: edit the PR title, this PR only remove some logs.
You wanted those logs, why do you remove them?

myakove (changes requested):

ocp_resources/virtual_machine_instance.py
vsibirsk: title is still relevant as the pr is to fix flakiness caused by this log
also your pr https://github.com/RedHatQE/openshift-python-wrapper/pull/1367/ added more logging that makes this one redundant

myakove (approved):

PR #1412, opened 2023-09-06
Update catalog_source class constructor, documentation and add 'to_dict' parameters validation

oharan2:
##### Short description:

Fix class constructor to not demand upper (super) classes args, fixed documentation


ocp_resources/catalog_source.py
```diff
@@ -1,34 +1,32 @@
...
-            yaml_file=yaml_file,
-            delete_timeout=delete_timeout,
-            **kwargs,
-        )
+        """
+        Args:
+            source_type (str): Name of the source type.
+            image (str): Image index for the catalog.
+            display_name (str): Display name for the catalog in the web console and CLI.
+            publisher (str): Name of the publisher.
```
rnetser, line 25: should raise if any of these are missing

rnetser (changes requested):

ocp_resources/catalog_source.py
oharan2: @rnetser Every pr is pointing at an issue like:
#1414 (Add validation for mandatory args for class constructor in 'catalog_source.py')

Can I fix it in this PR? (will change the pr title)
More resources will be updated similarly.

rnetser: i think that if you're already doing the job of adding docstrings and mapping what is mandatory, it will be easier to add the checks in the same pr; wdyt?

rnetser (approved):

PR #1422, opened 2023-09-06
Update configmap class constructor and documentation

oharan2:
Fix class constructor to not demand upper (super) classes args, fixed documentation


rnetser (changes requested):

ocp_resources/configmap.py
```diff
@@ -1,34 +1,23 @@
...
-            namespace=namespace,
-            client=client,
-            teardown=teardown,
-            yaml_file=yaml_file,
-            delete_timeout=delete_timeout,
-            **kwargs,
-        )
+        """
+        Args:
+            data (dict): key-value configuration pairs.
```
rnetser, line 18: should raise if missing (this is mandatory right?)

rnetser (changes requested):

ocp_resources/configmap.py
oharan2: Done

rnetser (approved):

PR #1444, opened 2023-09-11
Update controller_revision class constructor, documentation and add 'to_dict' parameters validation

oharan2:


ocp_resources/controller_revision.py
```diff
@@ -1,42 +1,36 @@
...
-            timeout=timeout,
-            privileged_client=privileged_client,
-            yaml_file=yaml_file,
-            delete_timeout=delete_timeout,
-            **kwargs,
-        )
+        """
+        Args:
+            owner_references (list): List of objects depended on this object.
+            revision_object (int): indicates the revision of the state represented by Data.
```
oharan2, line 20: @rnetser @myakove  I cannot verify this PR,
The changes I added are verified but the API docs mention the revision_object type is int, while line 39 is trying to interact with `revision_object.res` which doesn't match. What do you think?

[See](https://github.com/RedHatQE/openshift-python-wrapper/pull/1444/files#diff-3127d82cecc19690fb34f0c1c7bb69f195b28ee086bcbd3bc09e1a0e473aeed3R39)

rnetser: @oharan2 update `revision_object` description; it is not an int.
 `revision_object` is object spec.

rnetser (changes requested):

ocp_resources/controller_revision.py
oharan2: I thought the `revision_object` and the revision parameter is referring the same, will update

oharan2: @rnetser Do I need to validate the Int64 (as recently committed)? 
many cases might pass a regular int and I suggested to maintain this requirement.

pyproject.toml
```diff
@@ -64,3 +64,4 @@ python-benedict = "^0.32.0"
 packaging = "^23.1"
 python-simple-logger = "^1.0.6"
 jinja2 = "^3.1.2"
+numpy = "*"
```
myakove, line 67: Do not edit this file directly, use poetry add

ocp_resources/controller_revision.py
```diff
@@ -1,42 +1,42 @@
...
-            privileged_client=privileged_client,
-            yaml_file=yaml_file,
-            delete_timeout=delete_timeout,
-            **kwargs,
-        )
+        """
+        Args:
+            owner_references (list, optional): List of objects depended on this object.
+            revision_object (object, optional): the Data Object representing the state.
+            revision (int64): indicates the revision of the state represented by Data.
```
myakove, line 24: You specify that the code expect `int64` but in the code you convert it to `int64`
We need to get it from the user and only validate that the user was sent us `int64`

myakove (changes requested):

PR #1460, opened 2023-09-13
Add managedcluster resource and jinja2 dependency

rnetser:


pyproject.toml
```diff
@@ -63,3 +63,4 @@ kubernetes = "^28.1.0b1"
 python-benedict = "^0.32.0"
 packaging = "^23.1"
 python-simple-logger = "^1.0.6"
+jinja2 = "^3.1.2"
```
myakove, line 66: Why this is needed? Where it's used?

ocp_resources/managed_cluster.py
```diff
@@ -0,0 +1,10 @@
+from ocp_resources.resource import Resource
+
+
+class ManagedCluster(Resource):
+    """
+    https://github.com/stolostron/api/blob/main/cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml
+    Provides a representation of the managed cluster on the hub.
+    """
+
+    api_group = Resource.ApiGroup.CLUSTER_OPEN_CLUSTER_MANAGEMENT_IO
```
myakove, line 10: This is read only CRD? or the user can create one as well?

rnetser: readonly

pyproject.toml
rnetser: https://github.com/RedHatQE/openshift-python-wrapper/blob/main/tests/utils.py#L3

myakove (approved):

PR #1490, opened 2023-09-27
Update DataImportCron class constructor, documentation and add 'to_dict' parameters validation

oharan2:


ocp_resources/data_import_cron.py
```diff
@@ -57,8 +57,14 @@ def to_dict(self):
         super().to_dict()
         if not self.yaml_file:
             if self.image_stream and self.url:
-                raise ValueError("imageStream and url can not coexist")
-
+                raise ValueError("imageStream and url cannot coexist")
+            if not all(
+                [self.size, self.managed_data_source, self.pull_method, self.schedule]
```
myakove, line 62: `self.managed_data_source` and `self.schedule` marked as required in the API docs but have defaults so we cannot force the user to send them.

myakove: `self.size` is part of `storage` key which is not mandatory.

oharan2: Please checkout latest commits [approach](https://github.com/RedHatQE/openshift-python-wrapper/pull/1490/commits/2b9817657389968edf1c02c4a1de09c4252c4f2b#diff-bf49cfb5d402f8bb0e46e2125739760ec738f1a91689835bff8f8995f87beef1R99).
I can revert it in a second to the last usage of creating a storage on [L70](https://github.com/RedHatQE/openshift-python-wrapper/pull/1490/files#diff-bf49cfb5d402f8bb0e46e2125739760ec738f1a91689835bff8f8995f87beef1L70) and set also when `self.size` is None.

oharan2: Example output:
```
{'source': {'registry': {'pullMethod': '..', 'imageStream': '..'}}, 
'storage': {'resources': {'requests': {'storage': None}}, 'storageClassName': '..'}}
```

omrirh (approved):

ocp_resources/data_import_cron.py
```diff
@@ -57,21 +57,20 @@ def to_dict(self):
...
                     "spec": {
-                        "template": {
-                            "spec": {
-                                "source": {"registry": {"pullMethod": self.pull_method}},
-                                "storage": {"resources": {"requests": {"storage": self.size}}},
-                            }
-                        }
+                        "template": {"spec": {"source": {"registry": {"pullMethod": self.pull_method}}}},
+                        "managedDataSource": self.managed_data_source,
+                        "schedule": self.schedule,
```
rnetser, line 68: add only if provided

ocp_resources/data_import_cron.py
```diff
@@ -57,21 +57,20 @@ def to_dict(self):
...
                 {
                     "spec": {
-                        "template": {
-                            "spec": {
-                                "source": {"registry": {"pullMethod": self.pull_method}},
-                                "storage": {"resources": {"requests": {"storage": self.size}}},
-                            }
-                        }
+                        "template": {"spec": {"source": {"registry": {"pullMethod": self.pull_method}}}},
+                        "managedDataSource": self.managed_data_source,
```
rnetser, line 67: add only if provided

rnetser (changes requested):

rnetser (changes requested):
Comments were not addressed

rnetser (approved):

omrirh (approved):

myakove (approved):

PR #1499, opened 2023-10-05
WIP: Onboard ObservabilityAddon resource

omrirh:


ocp_resources/resource.py
```diff
@@ -317,6 +317,9 @@ class ApiGroup:
         V2V_KUBEVIRT_IO = "v2v.kubevirt.io"
         VELERO_IO = "velero.io"
         VM_KUBEVIRT_IO = "vm.kubevirt.io"
+        OBSERVABILITY_ADDON_IO = (
```
myakove, line 320: This should be sorted by ABC

ocp_resources/observability_addon.py
```diff
@@ -0,0 +1,9 @@
+from ocp_resources.resource import NamespacedResource
+
+
+class ObservabilityAddon(NamespacedResource):
```
myakove, line 4: Missing call to super and implementation on to_dict

myakove (changes requested):

ocp_resources/observability_addon.py
```diff
@@ -0,0 +1,9 @@
...
+    """
+    https://github.com/stolostron/multicluster-observability-operator/blob/main/docs/MultiClusterObservability-CRD.md
```
rnetser, line 6: is there an api doc?

ocp_resources/resource.py
rnetser, line 320: use `OBSERVABILITY_OPEN_CLUSTER_MANAGEMENT_IO` in f-string

rnetser (changes requested):

ocp_resources/observability_addon.py
omrirh: haven't found one yet; will update once I do

ocp_resources/resource.py
omrirh: done

ocp_resources/observability_addon.py
omrirh: I wish to use this resource for view only after the observability is enabled.
Should we implement a to_dict anyway?

myakove: Yes, this is not only for your personal use, this should be full resource implementation

PR #1559, opened 2023-12-28
Remove timeout sampler tests

myakove:
* Remove timeout sampler tests
* Remove pre-commit from tox


rnetser (changes requested):
we should also remove https://github.com/RedHatQE/openshift-python-wrapper/blob/main/tests/unittests/test_utils.py

myakove:
@rnetser I removed `unittests` folder completely

rnetser (approved):

PR #1583, opened 2024-01-12
WIP: Add DataScienceCluster resource for RHOAI operator

chetna14manku:


ocp_resources/data_science_cluster.py
```diff
@@ -0,0 +1,88 @@
...
+        Removed: The Operator actively manages the component but does not install it. If the component is
+        already installed, the Operator will try to remove it.
+        """
+
+        MANAGED = "Managed"
+        REMOVED = "Removed"
+
+    def __init__(
+        self,
+        name=None,
```
myakove, line 29: name, client and yaml_file should be removed (should be passed as part of `kwargs`)

ocp_resources/data_science_cluster.py
```diff
@@ -0,0 +1,88 @@
+from ocp_resources.resource import Resource
+
+
+class DataScienceCluster(Resource):
+    """
+    DataScienceCluster Resource for RHOAI operator
+
+    https://access.redhat.com/documentation/en-us/red_hat_openshift_ai_self-managed/2.5/html/installing_and_uninstalling_openshift_ai_self-managed/index
```
myakove, line 8: This is a link to `Installing and uninstalling OpenShift AI Self-Managed`, you need to replace it with link to the resource definition.

ocp_resources/resource.py
```diff
@@ -234,6 +234,7 @@ class ApiGroup:
         COORDINATION_K8S_IO = "coordination.k8s.io"
         DATA_IMPORT_CRON_TEMPLATE_KUBEVIRT_IO = "dataimportcrontemplate.kubevirt.io"
         DISCOVERY_K8S_IO = "discovery.k8s.io"
+        DATASCIENCE_CLUSTER = "datasciencecluster.opendatahub.io"
```
myakove, line 237: should be ordered alphabetic

myakove (changes requested):

ocp_resources/data_science_cluster.py
```diff
@@ -0,0 +1,88 @@
...
+        client=None,
+        yaml_file=None,
+        codeflare_state=ManagementState.REMOVED,
+        dashboard_state=ManagementState.MANAGED,
+        datasciencepipelines_state=ManagementState.REMOVED,
+        kserve_state=ManagementState.REMOVED,
+        modelmeshserving_state=ManagementState.REMOVED,
+        ray_state=ManagementState.REMOVED,
+        trustyai_state=ManagementState.REMOVED,
+        workbenches_state=ManagementState.REMOVED,
```
rnetser, line 39: do not set defaults; either the api has defaults and if the user does not provide - do not set in the code or they are mandatory and the user must provide them

rnetser (changes requested):

ocp_resources/data_science_cluster.py
chetna14manku: @myakove I didn't find any direct link to this resource, it is part of installation guide of data science product.

myakove: please add also `https://github.com/opendatahub-io/opendatahub-operator/blob/incubation/apis/datasciencecluster/v1/datasciencecluster_types.go`
and `https://github.com/opendatahub-io/opendatahub-operator/blob/a6eb73af559f9ba7438c529420f6e55c83b045af/config/crd/bases/datasciencecluster.opendatahub.io_datascienceclusters.yaml`

myakove (changes requested):

ocp_resources/data_science_cluster.py
chetna14manku: checking if there are defaults

ocp_resources/data_science_cluster.py
chetna14manku: Thanks, will update!

ocp_resources/data_science_cluster.py
chetna14manku: If there is no values passed for either one of the components, then DSC still getting created, without ManagemantState, see example. Removing the default values
```
spec:
  components:
    codeflare:
      devFlags: {}
      managementState: Removed
    dashboard:
      devFlags: {}
      managementState: Managed
    datasciencepipelines:
      devFlags: {}
      managementState: Removed
    kserve:
      devFlags: {}
      managementState: Removed
    modelmeshserving:
      devFlags: {}
      managementState: Removed
    ray:
      devFlags: {}
      managementState: Removed
    trustyai:
      devFlags: {}
      managementState: Removed
    workbenches:  ## >>>> NO State passed for this component <<<<<<
      devFlags: {}
```

chetna14manku:
moving back to WIP, adding DSC specific Conditions

chetna14manku:
Closing this PR, no need to add this resource.

PR #1587, opened 2024-01-14
Add snapshot property to Datasource class

hmeir:
##### Short description:
Datasource can have two types of bootsource: pvc and snapshot.
pvc property already exists, adding snapshot.

Adding snapshot property
refactor the "pvc" property so we access the boot source  according to the type requested
if the pvc/snapshot doesn't exists, return None

### DIFF
--- a/ocp_resources/data_source.py
+++ b/ocp_resources/data_source.py
@@ -3,6 +3,7 @@ from kubernetes.dynamic.exceptions import ResourceNotFoundError
 from ocp_resources.constants import TIMEOUT_4MINUTES
 from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
 from ocp_resources.resource import NamespacedResource
+from ocp_resources.volume_snapshot import VolumeSnapshot
 
 
 class DataSource(NamespacedResource):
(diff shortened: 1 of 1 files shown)


ocp_resources/data_source.py
```diff
@@ -39,19 +40,29 @@ def to_dict(self):
                 },
             })
 
-    @property
-    def pvc(self):
-        data_source_pvc = self.instance.spec.source.pvc
-        pvc_name = data_source_pvc.name
-        pvc_namespace = data_source_pvc.namespace
+    def _get_boot_source(self, boot_source_type):
```
myakove, line 43: You can get the resource class here directly.
and use it like: `self._get_boot_source(boot_source_type=PersistentVolumeClaim)`
No need for line 50 when calling it like this.

myakove (changes requested):

ocp_resources/data_source.py
hmeir: Hi ^^
Ye though about it,
but we need to get the boot source from the Datasource instance, so we will need to use "pvc" or "snapshot" like in line 44 in order to get its name & namespace

So we can all the function with the resource class like you suggested, but we will still need a condition for line 44 ,
and I found it easier to pass string to the function

wdyt?

ocp_resources/data_source.py
```diff
@@ -39,19 +40,29 @@ def to_dict(self):
...
+                f" {boot_source_name}, namespace: {boot_source_namespace}"
             )
+
+    @property
+    def pvc(self):
+        return self._get_boot_source(boot_source_type="pvc")
+
+    @property
+    def snapshot(self):
+        return self._get_boot_source(boot_source_type="snapshot")
```
rnetser, line 68: `boot_source` should be the property that returns either a PVC or snapshot

rnetser (changes requested):

ocp_resources/data_source.py
myakove: Missed that line, in this case what you doing here is OK.

ocp_resources/data_source.py
rnetser: To clarify - add `boot_source` and add deprecation on `pvc`

hmeir: Just to clarify:

You are suggesting that the only property should be `boot source`,
And that it should return PersisentVolumeClaim or VolumeSnapshot dynamically based on the the instance boot source type?

myakove: right, and add deprecation warning on `pvc` that it will be deprecated in 4.16.

hmeir: Hi

I have this PR in downstream: [https://code.engineering.redhat.com/gerrit/c/cnv-tests/+/449182](PR)
And I want to apply this DataSource change on it as well, because its accessing its snapshot/pvc references.
Its currently on main, but need to be backported to 4.15 & 4.14.

So If I change to `boot source` instead of `pvc`, the implementation will be different in 4.14,4.15.
So should I implement the downstream PR with `boot_source` and in 4.14&4.15 add the `snapshot` property like I did here?

PR #1605, opened 2024-01-23
Add StorageProfile properties

hmeir:
##### Short description:
Accessing StorageProfile status properties is very common, for example accessmode, volumemode, cloneStrategy etc.
Adding properties will allow easier access.
Also two function to get the first claimPropertySets accessmode & volume mode, very common in our tests as well.
In addition, Adding storage_profile property to StorageClass, because its very common to get the matching StorageProfile of the Storageclass.

Adding four common StorageProfile parameters as properties for easier Access.
In addition two functions for accessmode and volumemode.
Adding storage_profile property for StorageClass.

### DIFF
--- a/ocp_resources/storage_class.py
+++ b/ocp_resources/storage_class.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
+from kubernetes.dynamic.exceptions import ResourceNotFoundError
 
 from ocp_resources.resource import Resource
+from ocp_resources.storage_profile import StorageProfile
 
 
 class StorageClass(Resource):
@@ -102,3 +104,13 @@ class StorageClass(Resource):
                 self.res.update({"allowedTopologies": self.allowed_topologies})
             if self.mount_options:
                 self.res.update({"mountOptions": self.mount_options})
+
+    @property
+    def storage_profile(self):
+        try:
+            return StorageProfile(
+                client=self.client,
+                name=self.name,
+            )
+        except ResourceNotFoundError:
+            self.logger.warning(f" storageProfile is not found for {self.name}  storageClass")
--- a/ocp_resources/storage_class.py
+++ b/ocp_resources/storage_class.py
@@ -15,6 +15,9 @@ class StorageClass(Resource):
     class Types:
         """
         These are names of StorageClass instances when you run `oc get sc`
+
+        API:
+        https://kubernetes.io/docs/concepts/storage/storage-classes/
         """
 
         LOCAL_BLOCK = "local-block"
@@ -113,4 +116,5 @@ class StorageClass(Resource):
                 name=self.name,
             )
         except ResourceNotFoundError:
-            self.logger.warning(f" storageProfile is not found for {self.name}  storageClass")
+            self.logger.error(f" storageProfile is not found for {self.name}  storageClass")
+            raise
--- a/ocp_resources/storage_profile.py
+++ b/ocp_resources/storage_profile.py
@@ -4,4 +4,30 @@ from ocp_resources.resource import Resource
 
 
 class StorageProfile(Resource):
+    """
+    StorageProfile Object
+    """
+
     api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
+
+    def first_claim_property_set_access_modes(self):
+        return self.claim_property_sets[0]['accessModes']
+
+    def first_claim_property_set_volume_mode(self):
+        return self.claim_property_sets[0]['volumeMode']
+
+    @property
+    def claim_property_sets(self):
+        return self.instance.status.claimPropertySets
+
+    @property
+    def clone_strategy(self):
+        return self.instance.status.get("cloneStrategy")
+
+    @property
+    def data_import_cron_source_format(self):
+        return self.instance.status.get("dataImportCronSourceFormat")
+
+    @property
+    def snapshotclass(self):
+        return self.instance.status.get("snapshotClass")
--- a/ocp_resources/storage_profile.py
+++ b/ocp_resources/storage_profile.py
@@ -6,19 +6,22 @@ from ocp_resources.resource import Resource
 class StorageProfile(Resource):
     """
     StorageProfile Object
+
+    Doc:
+    https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md
     """
 
     api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
 
-    def first_claim_property_set_access_modes(self):
-        return self.claim_property_sets[0]['accessModes']
-
-    def first_claim_property_set_volume_mode(self):
-        return self.claim_property_sets[0]['volumeMode']
-
     @property
     def claim_property_sets(self):
-        return self.instance.status.claimPropertySets
+        return self.instance.status.get("claimPropertySets")
+
+    def first_claim_property_set_access_modes(self):
+        return self.claim_property_sets[0].get("volumeMode") if self.claim_property_sets else None
+
+    def first_claim_property_set_volume_mode(self):
+        return self.claim_property_sets[0].get("volumeMode") if self.claim_property_sets else None
 
     @property
     def clone_strategy(self):


ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,30 @@
 
 
 class StorageProfile(Resource):
+    """
```
rnetser, line 7: please add resource info according to https://github.com/RedHatQE/openshift-python-wrapper/blob/main/CONTRIBUTING.md#adding-a-new-module-resource

this is not a new resource but the info should be added to it

rnetser (changes requested):

ocp_resources/storage_profile.py
hmeir: Can a user create a custom storageprofile without creating storageclass?
I'm not sure if "init" fits here

rnetser: relationship is 1:1 for StorageClass : StorageProfile
so:
- when StorageClass is created, StorageProfile is also added
- adding StorageProfile without esixstingSC, it will be deleted
- adding StorageProfile with esixstingSC, it will be deleted

rnetser (approved):

ocp_resources/storage_class.py
```diff
@@ -102,3 +104,13 @@ def to_dict(self):
...
                 self.res.update({"mountOptions": self.mount_options})
+
+    @property
+    def storage_profile(self):
+        try:
+            return StorageProfile(
+                client=self.client,
+                name=self.name,
+            )
+        except ResourceNotFoundError:
```
rnetser, line 115: Based on what i understand, `StorageProfile` must exist for storage class. please raise if not found

rnetser (changes requested):

ocp_resources/storage_class.py
hmeir: done

rnetser (dismissed):

myakove (changes requested):

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,33 @@
...
+
     api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
+
+    @property
+    def claim_property_sets(self):
+        return self.instance.status.claimPropertySets
+
+    def first_claim_property_set_access_modes(self):
+        return self.claim_property_sets[0]["accessModes"]
```
rnetser, line 21: according to the doc
```
 If the storage provisioner defined in storage class does not have defaults configured in CDI the resulting StorageProfile has empty claimPropertySets.
```
this will fail if empty

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,33 @@
...
+
+    def first_claim_property_set_volume_mode(self):
+        return self.claim_property_sets[0]["volumeMode"]
```
rnetser, line 24: ditto

rnetser (changes requested):

ocp_resources/storage_profile.py
hmeir: Yes,
But in cnv-tests it used for StorageProfiles that we expect to have a claimPropertySets defined
So if its missing -> we expect the test to fail

If we just want to check/iterate over a storageprofile CPS, we have claim_property_sets property for it

rnetser: this repository is not tied to cnv-test; if you need some functionality that is not part of the api, please implement there.

rnetser (changes requested):

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,27 @@
...
+    """
+    StorageProfile Object
+
+    Doc:
+    https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md
+    """
+
     api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
+
+    @property
```
myakove, line 16: you shouldn't remove the functions, you need to return if exists, else none

myakove (changes requested):

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,33 @@
...
+    def claim_property_sets(self):
+        return self.instance.status.get("claimPropertySets")
+
+    def first_claim_property_set_access_modes(self):
+        return self.claim_property_sets[0].get("accessModes") if self.claim_property_sets else None
+
+    def first_claim_property_set_volume_mode(self):
+        return self.claim_property_sets[0].get("volumeMode") if self.claim_property_sets else None
```
myakove, line 24: I see that `self.claim_property_sets` is list, you should return the list and not list[0]

myakove (changes requested):

ocp_resources/storage_profile.py
hmeir: Its a list of objects
Each object has accessmodes & volumemode
And the property `claim_property_sets` returns a list already

The idea of this function is to return the accessmode/volumemode of the very first object in the list, usually for a  built-in StorageProfile that we expected for atleast one object to exist

In this scenario, if `self.claim_property_sets[0]` fails - then something is went wrong

myakove: ok, saw you rename the function.

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,33 @@
...
+
+    Doc:
+    https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md
+    """
+
     api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
+
+    @property
+    def claim_property_sets(self):
+        return self.instance.status.get("claimPropertySets")
```
myakove, line 18:
```suggestion
        return self.instance.status.get("claimPropertySets", {})
```

then you can remove `if self.claim_property_sets else None` from `first_claim_......`

myakove (changes requested):

ocp_resources/storage_profile.py
hmeir: So it will produce a KeyError if fails?
y nice

myakove (dismissed):

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,33 @@
...
+
+    Doc:
+    https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md
+    """
+
     api_group = Resource.ApiGroup.CDI_KUBEVIRT_IO
+
+    @property
+    def claim_property_sets(self):
+        return self.instance.status.get("claimPropertySets", {})
```
rnetser, line 18: why take it from `status` and  not `spec`? - applies to all places
`claimPropertySets` is a list not a dict

ocp_resources/storage_profile.py
```diff
@@ -4,4 +4,33 @@
...
+
+    def first_claim_property_set_access_modes(self):
+        return self.claim_property_sets[0].get("accessModes")
+
+    def first_claim_property_set_volume_mode(self):
+        return self.claim_property_sets[0].get("volumeMode")
```
rnetser, line 24: if `claim_property_sets` is empty, this will fail


```
In [1]: x = []

In [2]: x[0]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 x[0]

IndexError: list index out of range

```

rnetser (changes requested):

ocp_resources/storage_profile.py
hmeir: Because for the common case the `Spec` will be empty

When you create storgeclass the storageprofile is created automatically and its status params generated according to the storageclass defined
Spec is mostly to override/modify some params, but it will be updated in the status as well

So getting the params from the status is the right approach

myakove: `claimPropertySets is a list not a dict`


right, my bad (I gave example with dict instead of list)

ocp_resources/storage_profile.py
myakove: Yes, needs to be reverted to the old code.

PR #1641, opened 2024-02-02
create AAQ resource

dbasunag:

### DIFF
--- a/ocp_resources/aaq.py
+++ b/ocp_resources/aaq.py
@@ -1,5 +1,6 @@
 # API reference: https://github.com/kubevirt/application-aware-quota/blob/main/README.md
 # TODO: update API reference when OCP doc is available
+# TODO: Add missing code e.g. to_dict()
 
 from ocp_resources.resource import Resource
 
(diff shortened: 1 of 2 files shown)


ocp_resources/aaq.py
```diff
@@ -0,0 +1,12 @@
+# API reference: https://github.com/kubevirt/application-aware-quota/blob/main/README.md
+# TODO: update API reference when OCP doc is available
+
+from ocp_resources.resource import Resource
+
+
+class AAQ(Resource):
```
myakove, line 7: Is this resource that the user can create as well?

dbasunag: I believe so - this work is going to evolve and provide more clarity in next few months. For our use case, it's deployment is currently controlled via HCO and this support was added just now.

myakove: Please add TODO, to fill the missing code for create (to_dict())

myakove (changes requested):

ocp_resources/aaq.py
dbasunag: Added one todo.

rnetser (approved):

myakove (approved):

PR #1705, opened 2024-03-05
Fix storageProfile access_modes property

hmeir:
Right now in StorageProfile class, the function **first_claim_property_set_access_modes** getting the "volumeMode" instead of "accessModes".
So this PR is fixing it.

### DIFF
--- a/ocp_resources/storage_profile.py
+++ b/ocp_resources/storage_profile.py
@@ -18,7 +18,7 @@ class StorageProfile(Resource):
         return self.instance.status.get("claimPropertySets")
 
     def first_claim_property_set_access_modes(self):
-        return self.claim_property_sets[0].get("volumeMode") if self.claim_property_sets else None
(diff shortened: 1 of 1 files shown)


myakove (approved):

redhat-qe-bot1:
Cherry-picked PR Fix storageProfile access_modes property into v4.15

ocp_resources/storage_profile.py
```diff
@@ -18,7 +18,7 @@ def claim_property_sets(self):
         return self.instance.status.get("claimPropertySets")
 
     def first_claim_property_set_access_modes(self):
-        return self.claim_property_sets[0].get("volumeMode") if self.claim_property_sets else None
+        return self.claim_property_sets[0].get("accessModes") if self.claim_property_sets else None
```
jpeimer, line 21: If we want to get the **first** access mode, shouldn't it be
`claim_property_sets[0].get("accessModes")[0]`
?

Issue #1543, opened 2023-12-13
Allow python wrapper to accept tokenrequest api that are not enabled under API resources


geetikakay:
**Is your feature request related to a problem? Please describe.**
Support for api like Token request which is not a valid API that can be enabled under API resources. The API is a subresource of the service account API, see the docs for more information on the structure and the operations available for this API, https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/

**Describe the solution you'd like**
In golang we are able to use it but with wrapper when i tried it failed as it's not part of cluster's api-resources.
Refer : https://issues.redhat.com/browse/OCPBUGS-22913

**Describe alternatives you've considered**
I thought of three ways 
1. probably add support in openshift-python wrapper
2. in d/s code, we can use python request and send kubeadmin token to sign other tokens.
3. use third party client tool: oc create token 

**Additional context**

PR #1777, opened 2024-04-15
Allow token creation for service accounts

dbasunag:
Most likely addresses https://github.com/RedHatQE/openshift-python-wrapper/issues/1543


ocp_resources/service_account.py
```diff
@@ -1,9 +1,52 @@
...
+        super().to_dict()
+        if not self.yaml_file:
+            if self.automount_service_account_token:
+                self.res["automountServiceAccountToken"] = self.automount_service_account_token
+            if self.image_pull_secrets:
+                self.res["imagePullSecrets"] = self.image_pull_secrets
+            if self.secrets:
+                self.res["secrets"] = self.secrets
+
+    def create_service_account_token(self, expiration_seconds=None, audiences=None, bound_object_ref=None):
```
rnetser, line 41: please add docstring
Can each of the args be None and still passed?

based on https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1TokenRequestSpec.md, `audiences` is mandatory

rnetser (changes requested):

ocp_resources/service_account.py
dbasunag: It uses defaults. ```>>> out = sa.create_service_account_token(expiration_seconds=600)
>>> print(out)
{'api_version': 'authentication.k8s.io/v1', 'kind': 'TokenRequest', 'metadata': {'annotations': None, 'creation_timestamp': datetime.datetime(2024, 4, 15, 22, 21, 25, tzinfo=tzutc()), 'deletion_grace_period_seconds': None, 'deletion_timestamp': None, 'finalizers': None, 'generate_name': None, 'generation': None, 'labels': None, 'managed_fields': [{'api_version': 'authentication.k8s.io/v1', 'fields_type': 'FieldsV1', 'fields_v1': {'f:spec': {'f:expirationSeconds': {}}}, 'manager': 'OpenAPI-Generator', 'operation': 'Update', 'subresource': 'token', 'time': datetime.datetime(2024, 4, 15, 22, 21, 25, tzinfo=tzutc())}], 'name': 'prometheus-k8s', 'namespace': 'openshift-monitoring', 'owner_references': None, 'resource_version': None, 'self_link': None, 'uid': None}, 'spec': {'audiences': ['https://kubernetes.default.svc'], 'bound_object_ref': None, 'expiration_seconds': 600}, 'status': {'expiration_timestamp': datetime.datetime(2024, 4, 15, 22, 31, 25, tzinfo=tzutc()), 'token': '<token_string>'}}
```

File: ocp_resources/service_account.py [hunk 2]
```diff
@@ -1,9 +1,59 @@
...
+
+    def create_service_account_token(self, audiences=None, bound_object_ref=None, expiration_seconds=None):
+        """
+        Args:
+            audiences (list, Optional): Intended audiences of the token
+            bound_object_ref (BoundObjectReference, Optional): Reference to an object that a token is bound to
+            expiration_seconds (int, Optional): Requested duration of validity of the token in seconds.
+                                                Defaults to 3600 seconds
+        """
+        return self._kube_v1_api.create_namespaced_service_account_token(
```
myakove, line 49: `TokenRequestSpec` is part of `TokenRequest` https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/

Please create a new resource and use it

myakove (changes requested):

ocp_resources/service_account.py
myakove: This is not optional.

https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/

dbasunag: This is using https://docs.openshift.com/container-platform/4.15/authentication/bound-service-account-tokens.html, with the kubernetes default 'https://kubernetes.default.svc'. This is the reason, create_service_account_token() should not make audiences as a required param.

ocp_resources/service_account.py
dbasunag: Ok. I need to use https://docs.openshift.com/container-platform/4.15/authentication/bound-service-account-tokens.html, with only validation present being spec is not `None` (which the current code gives me). As per https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/ I would need to make audiences required param, if a resource is created - I would need to avoid that. Please advice.

ocp_resources/service_account.py
dbasunag: As discussed not using TokenRequest resource, as it is not a api-resource.

ocp_resources/service_account.py
```diff
@@ -1,9 +1,59 @@
...
+            if self.secrets:
+                self.res["secrets"] = self.secrets
+
+    def create_service_account_token(self, audiences=None, bound_object_ref=None, expiration_seconds=None):
+        """
+        Args:
+            audiences (list, Optional): Intended audiences of the token
+            bound_object_ref (BoundObjectReference, Optional): Reference to an object that a token is bound to
+            expiration_seconds (int, Optional): Requested duration of validity of the token in seconds.
+                                                Defaults to 3600 seconds
```
rnetser, line 47: please remove; this is the api's curent default, we should not mention things that are not set in the code here

ocp_resources/service_account.py
```diff
@@ -1,9 +1,59 @@
...
+            name=self.name,
+            namespace=self.namespace,
+            body=kubernetes.client.AuthenticationV1TokenRequest(
+                spec={
+                    "audiences": audiences,
+                    "boundObjectRef": bound_object_ref,
+                    "expirationSeconds": expiration_seconds,
+                }
+            ),
+        ).to_dict()
```
rnetser, line 59: please remove `to_dict` - return the object; whoever calls this function can do it, if needed

rnetser (changes requested):

ocp_resources/service_account.py
```diff
@@ -1,9 +1,59 @@
...
+            if self.image_pull_secrets:
+                self.res["imagePullSecrets"] = self.image_pull_secrets
+            if self.secrets:
+                self.res["secrets"] = self.secrets
+
+    def create_service_account_token(self, audiences=None, bound_object_ref=None, expiration_seconds=None):
+        """
+        Args:
+            audiences (list, Optional): Intended audiences of the token
+            bound_object_ref (BoundObjectReference, Optional): Reference to an object that a token is bound to
```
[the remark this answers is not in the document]
dbasunag: It would be a dict: https://github.com/kubernetes-client/python/blob/e0234d3e7193aea4b4e8f487cab0f1228f4386d2/kubernetes/docs/V1BoundObjectReference.md

rnetser (approved):

ocp_resources/service_account.py
dbasunag: Done

myakove (approved):

PR #1697, opened 2024-03-04
Add vlan and mtu parameters to OVNOverlayNetworkAttachmentDefinition

Anatw:
##### Short description:
Add vlan and mtu parameters to OVNOverlayNetworkAttachmentDefinition.

##### What this PR does / why we need it:
Needed for the localnet feature.

PR #1781, opened 2024-04-19
Fix a mix between spec.netAttachDefName and spec.name fields

Anatw:

### DIFF
--- a/ocp_resources/network_attachment_definition.py
+++ b/ocp_resources/network_attachment_definition.py
@@ -214,8 +214,8 @@ class OVNOverlayNetworkAttachmentDefinition(NetworkAttachmentDefinition):
 
         Args:
             topology (str): The secondary network topology to be created.
-            network_name (str, optional): The name of the network, including the namespace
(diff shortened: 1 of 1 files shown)


myakove (commented):
This means that after your last PR this was broken? 
If so, how did you verify your last PR?

Anatw:
> This means that after your last PR this was broken? If so, how did you verify your last PR?

I verified it by running an entire module creating such resources. I am not sure how this happened,  I might have accidentally ran it against master and not the PR.

segevyos (commented):
/lgtm

myakove:
your original PR we reverted.

segevyos:
> your original PR we reverted.

@myakove The [original PR](https://github.com/RedHatQE/openshift-python-wrapper/pull/1697) also included VLAN and MTU, so reverting it also removes the support of these parameters.

myakove:
The original PR breaks the code, therefore needs to be reverted

PR #1863, opened 2024-05-30
Support for SelfNodeRemediationTemplate

geetikakay:
##### Short description:
Support  for SelfNodeRemediation resources for create  SelfNodeRemediation CRD's 

This PR is to support default remediation for NodeHealthCheck which is  SelfNodeRemediation

### DIFF
--- a/ocp_resources/self_node_remediation_templates.py
+++ b/ocp_resources/self_node_remediation_templates.py
@@ -1,0 +1,36 @@
+# API reference: https://github.com/medik8s/self-node-remediation
+from typing import Any
+
+from ocp_resources.resource import NamespacedResource
+
+
+class SelfNodeRemediationTemplate(NamespacedResource):
+    """
+    SelfNodeRemediationTemplate CRD
+    """
+
+    api_group = NamespacedResource.ApiGroup.SELF_NODE_REMEDIATION_MEDIK8S_IO
+
+    def __init__(
+        self,
+        remediation_strategy: str = "",
+        **kwargs: Any,
+    ) -> None:
+        """
+        Create SelfNodeRemediationTemplate object.
+
+        Args:
+            remediation_strategy (str, optional): The remediation method for unhealthy nodes.
+        """
+        super().__init__(
+            **kwargs,
+        )
+        self.remediation_strategy = remediation_strategy
+
+    def to_dict(self) -> None:
+        super().to_dict()
+        if not self.yaml_file:
+            _spec = self.res["spec"] = {"template": {"spec": {}}}
+
+            if self.remediation_strategy:
+                _spec["template"]["spec"]["remediationStrategy"] = self.remediation_strategy
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -325,6 +325,7 @@ class Resource:
         SCHEDULING_K8S_IO = "scheduling.k8s.io"
         SECURITY_ISTIO_IO = "security.istio.io"
         SECURITY_OPENSHIFT_IO = "security.openshift.io"
+        SELF_NODE_REMEDIATION_MEDIK8S_IO = "self-node-remediation.medik8s.io"
         SNAPSHOT_STORAGE_K8S_IO = "snapshot.storage.k8s.io"
         SNAPSHOT_KUBEVIRT_IO = "snapshot.kubevirt.io"
         SRIOVNETWORK_OPENSHIFT_IO = "sriovnetwork.openshift.io"


ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,47 @@
...
+        Args:
+            remediation_strategy (str, optional): The remediation method for unhealthy nodes.
+                Options are "Automatic", "OutOfServiceTaint", or "ResourceDeletion".
+                Default is "Automatic".
+        """
+        if remediation_strategy not in SelfNodeRemediationTemplate.VALID_STRATEGIES:
+            raise ValueError(
+                f"Invalid remediation strategy: {remediation_strategy}. "
+                f"Supported values are: {', '.join(SelfNodeRemediationTemplate.VALID_STRATEGIES)}"
+            )
```
rnetser, line 32: should be done by the resource api; please remove

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    SelfNodeRemediationTemplate CRD
+    """
+
+    api_group = NamespacedResource.ApiGroup.SELF_NODE_REMEDIATION_MEDIK8S_IO
+    VALID_STRATEGIES = {"Automatic", "ResourceDeletion", "OutOfServiceTaint"}
+
+    def __init__(
+        self,
+        remediation_strategy: Union[str, None] = "Automatic",
```
rnetser, line 17: do not set a default; the user should pass the desired strategy

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,47 @@
...
+            **kwargs,
+        )
+        self.remediation_strategy = remediation_strategy
+
+    def to_dict(self) -> None:
+        super().to_dict()
+        if not self.yaml_file:
+            spec = self.res.setdefault("spec", {})
+            spec["template"] = spec.get("template", {})
+            spec["template"]["spec"] = spec["template"].get("spec", {})
```
rnetser, line 44: please check how this is done in the examples referenced in the contribution doc

rnetser (changes requested):

ocp_resources/self_node_remediation_templates.py
geetikakay: ack, Thanks

rnetser (changes requested):

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,39 @@
...
+class SelfNodeRemediationTemplate(NamespacedResource):
+    """
+    SelfNodeRemediationTemplate CRD
+    """
+
+    api_group = NamespacedResource.ApiGroup.SELF_NODE_REMEDIATION_MEDIK8S_IO
+
+    def __init__(
+        self,
+        remediation_strategy: Union[str, None] = None,
```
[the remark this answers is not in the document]
geetikakay: Sure, I think this should change. I was going over the code : https://github.com/medik8s/self-node-remediation/blob/main/api/v1alpha1/selfnoderemediation_types.go#L49, it uses omitempty for this field.

maybe remediation_strategy: Optional[str] = None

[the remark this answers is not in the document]
rnetser: `remediation_strategy: str = ""` and set `if self.remediation_strategy`

rnetser (changes requested):

omrirh (approved):

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+        self.remediation_strategy = remediation_strategy
+
+    def to_dict(self) -> None:
+        super().to_dict()
+        if not self.yaml_file:
+            spec = self.res.setdefault("spec", {}).setdefault("template", {}).setdefault("spec", {})
+
+            if self.remediation_strategy:
+                spec["remediationStrategy"] = self.remediation_strategy
```
myakove, line 36:
```suggestion
                self.res["spec"] = {"template": "spec": {"remediationStrategy": self.remediation_strategy}}
```

myakove (changes requested):

ocp_resources/self_node_remediation_templates.py
geetikakay: @myakove not sure why we can't go with below or already existing one. 
spec is optional but under that we have template.spec which are mandatory 

            spec = self.res.setdefault("spec", {}).setdefault("template", {}).setdefault("spec", {})

            if self.remediation_strategy:
                spec["remediationStrategy"] = self.remediation_strategy

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        super().__init__(
+            **kwargs,
+        )
+        self.remediation_strategy = remediation_strategy
+
+    def to_dict(self) -> None:
+        super().to_dict()
+        if not self.yaml_file:
+            spec = self.res.setdefault("spec", {}).setdefault("template", {}).setdefault("spec", {})
```
myakove, line 33: not used

myakove (changes requested):

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        super().__init__(
+            **kwargs,
+        )
+        self.remediation_strategy = remediation_strategy
+
+    def to_dict(self) -> None:
+        super().to_dict()
+        if not self.yaml_file:
+            self.res.setdefault("spec", {}).setdefault("template", {}).setdefault("spec", {})
```
geetikakay, line 33: @myakove Hi, by doing  self.res.setdefault("spec", {}).setdefault("template", {}).setdefault("spec", {}) i am making sure that i have { "spec": { "template": { "spec": {} } } } available because template,spec all are required fields . 

if remediation_strategy is there maybe just adding value self.res["spec"]["template"]["spec"]["remediationStrategy"] = self.remediation_strategy would do instead of direct replacement.

myakove: you do not need to make sure of it since you set it as literal
`{"template": {"spec": {"remediationStrategy": self.remediation_strategy}}}`

geetikakay: self.res.setdefault("spec", {}).setdefault("template", {}).setdefault("spec", {}) is mandatory to have even if remediationStrategy is not set else we will hit this [problem](https://issues.redhat.com/browse/ECOPROJECT-1999)

myakove: ok, then write it as literal, no need to use `setdefault`

myakove (changes requested):

ocp_resources/self_node_remediation_templates.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        super().__init__(
+            **kwargs,
+        )
+        self.remediation_strategy = remediation_strategy
+
+    def to_dict(self) -> None:
+        super().to_dict()
+        if not self.yaml_file:
+            self.res.update({"spec": {"template": {"spec": {}}}})
```
myakove, line 33: we do not update or setdefault when create resource.
we set it from scratch.

```
_spec = self.res["spec"] = {"template": {"spec": {}}}
```
Then use `_spec`

myakove (changes requested):

myakove (approved):

PR #1884, opened 2024-06-21
Add TrustyAIService resource

adolfo-ab:
##### Short description:
Adds the TrustyAIService resource

##### More details:
This resource allows users to deploy TrustyAIService, which provides explainability features for AI models

### DIFF
--- a/ocp_resources/trustyai_service.py
+++ b/ocp_resources/trustyai_service.py
@@ -18,7 +18,7 @@ class TrustyAIService(NamespacedResource):
         image: Optional[str] = None,
         tag: Optional[str] = None,
         **kwargs,
-    ):
+    ) -> None:
         """
         TrustyAIService object
 
--- a/ocp_resources/trustyai_service.py
+++ b/ocp_resources/trustyai_service.py
@@ -1,0 +1,59 @@
+from ocp_resources.resource import NamespacedResource
+from typing import Optional, Any, Dict
+
+
+class TrustyAIService(NamespacedResource):
+    """
+    https://github.com/trustyai-explainability/trustyai-service-operator/blob/main/api/v1alpha1/trustyaiservice_types.go
+    """
+
+    api_group: str = NamespacedResource.ApiGroup.TRUSTYAI_OPENDATAHUB_IO
+
+    def __init__(
+        self,
+        storage: Dict[str, Any],
+        data: Dict[str, Any],
+        metrics: Dict[str, Any],
+        replicas: Optional[int] = 1,
+        image: Optional[str] = None,
+        tag: Optional[str] = None,
+        **kwargs,
+    ):
+        """
+        TrustyAIService object
+
+        Args:
+            storage (Dic[str, Any]): TrustyAIService storage config (format, folder and size).
+            data (Dict[str, Any]): TrustyAIService data config (filename and format).
+            metrics (Dict[str, Any]): TrustyAIService metrics config.
+            replicas (Optional[int], default: 1): Number of replicas for the TrustyAIService.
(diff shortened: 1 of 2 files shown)


ocp_resources/trustyai_service.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    api_group: str = NamespacedResource.ApiGroup.TRUSTYAI_OPENDATAHUB_IO
+
+    def __init__(
+        self,
+        storage: Dict[str, Any],
+        data: Dict[str, Any],
+        metrics: Dict[str, Any],
+        replicas: Optional[int] = 1,
```
myakove, line 17: Please do not set any defaults, if the user sent we set it, else we let k8s to handle it and set the default.

ocp_resources/trustyai_service.py
```diff
@@ -0,0 +1,59 @@
...
+class TrustyAIService(NamespacedResource):
+    """
+    https://github.com/trustyai-explainability/trustyai-service-operator/blob/main/api/v1alpha1/trustyaiservice_types.go
+    """
+
+    api_group: str = NamespacedResource.ApiGroup.TRUSTYAI_OPENDATAHUB_IO
+
+    def __init__(
+        self,
+        storage: Dict[str, Any],
```
myakove, line 14: all should be Optional, the user can pass yaml file instead.

ocp_resources/trustyai_service.py
```diff
@@ -0,0 +1,59 @@
...
+        self.data = data
+        self.metrics = metrics
+        self.replicas = replicas
+        self.image = image
+        self.tag = tag
+
+    def to_dict(self) -> None:
+        super().to_dict()
+
+        if not self.yaml_file:
```
myakove, line 44: here you can raise if some mandatory args are not passed.

myakove (changes requested):

ocp_resources/trustyai_service.py
```diff
@@ -0,0 +1,62 @@
...
+        **kwargs,
+    ) -> None:
+        """
+        TrustyAIService object
+
+        Args:
+            storage (Optional[Dic[str, Any]]): TrustyAIService storage config (format, folder and size).
+            data (Optional[Dict[str, Any]]): TrustyAIService data config (filename and format).
+            metrics (Optional[Dict[str, Any]]): TrustyAIService metrics config.
+            replicas (Optional[int], default: 1): Number of replicas for the TrustyAIService.
```
rnetser, line 29: please remove the default value as it is not set in the wrappers' code but by the api

ocp_resources/trustyai_service.py
```diff
@@ -0,0 +1,62 @@
...
+        self.replicas = replicas
+        self.image = image
+        self.tag = tag
+
+    def to_dict(self) -> None:
+        super().to_dict()
+
+        if not self.yaml_file:
+            if not (self.storage or self.data or self.metrics):
+                raise MissingRequiredArgumentError(argument="'storage' or 'data' or 'metrics'")
```
rnetser, line 46: please remove the `optional` from the docstring for these attributes

rnetser (changes requested):

ocp_resources/trustyai_service.py
myakove: They are still optional from the function perspective and should stay as optional in the docsting.

rnetser: so `mandatory if not passing yaml_file` should be added to the relevant places in the docstring

myakove: ok

myakove (approved):

PR #1889, opened 2024-07-04
Fix race in resource delete

jpeimer:
##### Short description:
- `client_wait_deleted` now returns bool
- `wait_deleted` now returns bool
- in resource `delete`: moving the get `hashed_data` under `try/except`.

##### Reason:
When the object is in Terminating state, it passes the check` if self.exists:`, but then NotFoundError gets raised on `(resource_dict=self.instance...` because the object is already gone.

For example, when we delete a DV, PVC gets deleted automatically.

When some automation attempts to create a DV twice, and then attempts to delete this DV twice, we may hit this race.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -806,18 +806,18 @@ class Resource:
     def delete(self, wait: bool = False, timeout: int = TIMEOUT_4MINUTES, body: Dict[str, Any] | None = None) -> bool:
         self.logger.info(f"Delete {self.kind} {self.name}")
 
-        try:
-            if self.exists:
+        if self.exists:
+            try:
                 hashed_data = self.hash_resource_dict(resource_dict=self.instance.to_dict())
                 self.logger.info(f"Deleting {hashed_data}")
                 self.logger.debug(f"\n{yaml.dump(hashed_data)}")
 
-            self.api.delete(name=self.name, namespace=self.namespace, body=body)
-            if wait:
-                self.client_wait_deleted(timeout=timeout)
+                self.api.delete(name=self.name, namespace=self.namespace, body=body)
+                if wait:
+                    self.client_wait_deleted(timeout=timeout)
 
-        except (NotFoundError, TimeoutExpiredError):
-            return False
+            except (NotFoundError, TimeoutExpiredError):
+                return False
 
         return True
 
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -806,12 +806,12 @@ class Resource:
     def delete(self, wait: bool = False, timeout: int = TIMEOUT_4MINUTES, body: Dict[str, Any] | None = None) -> bool:
         self.logger.info(f"Delete {self.kind} {self.name}")
 
-        if self.exists:
-            hashed_data = self.hash_resource_dict(resource_dict=self.instance.to_dict())
-            self.logger.info(f"Deleting {hashed_data}")
-            self.logger.debug(f"\n{yaml.dump(hashed_data)}")
-
         try:
+            if self.exists:
+                hashed_data = self.hash_resource_dict(resource_dict=self.instance.to_dict())
+                self.logger.info(f"Deleting {hashed_data}")
+                self.logger.debug(f"\n{yaml.dump(hashed_data)}")
+
             self.api.delete(name=self.name, namespace=self.namespace, body=body)
             if wait:
                 self.client_wait_deleted(timeout=timeout)
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -818,6 +818,8 @@ class Resource:
 
             except (NotFoundError, TimeoutExpiredError):
                 return False
+        else:
+            LOGGER.warning(f"Resource {self.kind} {self.name} does not exist")
 
         return True
 
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -679,7 +679,7 @@ class Resource:
             if sample:
                 return
 
-    def wait_deleted(self, timeout: int = TIMEOUT_4MINUTES) -> None:
+    def wait_deleted(self, timeout: int = TIMEOUT_4MINUTES) -> bool:
         """
         Wait until resource is deleted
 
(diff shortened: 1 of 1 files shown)


omrirh (approved):

ocp_resources/resource.py
```diff
@@ -806,12 +806,12 @@ def create(self, wait: bool = False) -> Optional[ResourceInstance]:
     def delete(self, wait: bool = False, timeout: int = TIMEOUT_4MINUTES, body: Dict[str, Any] | None = None) -> bool:
         self.logger.info(f"Delete {self.kind} {self.name}")
 
-        if self.exists:
-            hashed_data = self.hash_resource_dict(resource_dict=self.instance.to_dict())
-            self.logger.info(f"Deleting {hashed_data}")
-            self.logger.debug(f"\n{yaml.dump(hashed_data)}")
-
         try:
+            if self.exists:
```
myakove, line 810: This code needs to be changed

We should return False if not exists and log a proper error that the resource do not exist.
We need to call delete only if the resource exists.
There is no race, if you want to make sure the resource deleted pass wait=True.

myakove (changes requested):

ocp_resources/resource.py
jpeimer: At this moment, resource still exists. It's gone less than a second after this check, so the `instance` call fails.

myakove (changes requested):

ocp_resources/resource.py
jpeimer: > Do you mean that in line 809 the resource exists but in line 810 it's already gone?
Yes, exactly
Sure, will move it, thanks!

ocp_resources/resource.py
```diff
@@ -807,17 +807,17 @@ def delete(self, wait: bool = False, timeout: int = TIMEOUT_4MINUTES, body: Dict
...
+            try:
+                hashed_data = self.hash_resource_dict(resource_dict=self.instance.to_dict())
+                self.logger.info(f"Deleting {hashed_data}")
+                self.logger.debug(f"\n{yaml.dump(hashed_data)}")
 
-        try:
-            self.api.delete(name=self.name, namespace=self.namespace, body=body)
-            if wait:
-                self.client_wait_deleted(timeout=timeout)
+                self.api.delete(name=self.name, namespace=self.namespace, body=body)
```
dbasunag, line 815: If someone calls delete, we expect the resource to exist and not be automatically deleted by the product design. This change would hide such problems. Is that intentional?

ocp_resources/resource.py
dbasunag: A resource can exist in line 809 but be gone in 810, if it was marked for automatic deletion, not as a part of delete() call. Should delete() call support deletion of such resources?

myakove: delete() should act as best effort, if the resource is there when we called we will try to delete it, if the resource is gone while we try we return True (the user wanted the resource to be delete, it was delete even if it not by us)

ocp_resources/resource.py
```diff
@@ -807,17 +807,17 @@ def delete(self, wait: bool = False, timeout: int = TIMEOUT_4MINUTES, body: Dict
...
+                if wait:
+                    self.client_wait_deleted(timeout=timeout)
 
-        except (NotFoundError, TimeoutExpiredError):
-            return False
+            except (NotFoundError, TimeoutExpiredError):
+                return False
```
myakove, line 821: Please add warn log when the resource is not exists.

myakove (changes requested):

ocp_resources/resource.py
jpeimer: Do you mean if not `exists`, or if we got an exception?

myakove: if not exists, the user needs to know that we do not tried to delete it since it wan not exists when called.

ocp_resources/resource.py
```diff
@@ -807,17 +807,19 @@ def delete(self, wait: bool = False, timeout: int = TIMEOUT_4MINUTES, body: Dict
...
+                self.logger.info(f"Deleting {hashed_data}")
+                self.logger.debug(f"\n{yaml.dump(hashed_data)}")
 
-        try:
-            self.api.delete(name=self.name, namespace=self.namespace, body=body)
-            if wait:
-                self.client_wait_deleted(timeout=timeout)
+                self.api.delete(name=self.name, namespace=self.namespace, body=body)
+                if wait:
+                    self.client_wait_deleted(timeout=timeout)
```
myakove, line 817: please return `self.client_wait_deleted(timeout=timeout)` (may need to change `client_wait_deleted` return values)

myakove (changes requested):

ocp_resources/resource.py
jpeimer: Done

myakove (approved):

ocp_resources/resource.py
jpeimer: Done

myakove (approved):

redhat-qe-bot2:
Cherry-picked PR Fix race in resource delete into v4.16

msi-b1:
**Manual cherry-pick is needed**
Cherry pick failed for 60d585cdf63291112dcaf3a06b00507ee13caffc to v4.15:
To cherry-pick run:
```
git remote update
git checkout v4.15
git pull origin v4.15
git checkout -b fix_race_in_delete-v4.15
git cherry-pick 60d585cdf63291112dcaf3a06b00507ee13caffc
git push origin fix_race_in_delete-v4.15
```

redhat-qe-bot2:
**Manual cherry-pick is needed**
Cherry pick failed for 60d585cdf63291112dcaf3a06b00507ee13caffc to v4.14:
To cherry-pick run:
```
git remote update
git checkout v4.14
git pull origin v4.14
git checkout -b fix_race_in_delete-v4.14
git cherry-pick 60d585cdf63291112dcaf3a06b00507ee13caffc
git push origin fix_race_in_delete-v4.14
```

PR #1935, opened 2024-07-25
api version - use class attr and extract api group from api version

rnetser:


scripts/resource/class_generator.py
```diff
@@ -278,13 +278,33 @@ def parse_explain(
 
     LOGGER.debug(f"\n{yaml.dump(resource_dict)}")
 
-    if api_group_real_name := resource_dict.get("GROUP"):
+    # If API Group is not present in resource, try to get it from VERSION
+    if not (api_group_real_name := resource_dict.get("GROUP")) and (
```
myakove, line 282: please do not use := on negative condition, it's not clear.

myakove (changes requested):

myakove (approved):

myakove (approved):

myakove (approved):

PR #2066, opened 2024-08-22
Add support to set client with cluster proxy

sbahar619:
##### Short description:
Add support to set client with cluster proxy
https://issues.redhat.com/browse/CNV-46351
## Summary by CodeRabbit

- **New Features**
	- Enhanced Kubernetes client configuration to support proxy settings, improving flexibility for users connecting through proxies.

- **Bug Fixes**
	- Ensured proper logging of proxy settings for better transparency and troubleshooting.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -13,7 +13,7 @@ from types import TracebackType
 from typing import Optional, Any, Dict, List
 
 import kubernetes
-from kubernetes import config, client
+from kubernetes import client
 from kubernetes.dynamic import DynamicClient, ResourceInstance
 import yaml
 from benedict import benedict
@@ -79,7 +79,11 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
 
 
 def get_client(
-    config_file: str = "", config_dict: Dict[str, Any] | None = None, context: str = "", proxy: str = "", **kwargs: Any
+    config_file: str = "",
+    config_dict: Dict[str, Any] | None = None,
+    context: str = "",
+    proxy: Optional[str] = "",
+    **kwargs: Any,
 ) -> DynamicClient:
     """
     Get a kubernetes client.
@@ -124,7 +128,9 @@ def get_client(
         LOGGER.info("Trying to get client via new_client_from_config")
 
         return kubernetes.dynamic.DynamicClient(
-            client=kubernetes.config.new_client_from_config(config_file=config_file, client_configuration=client_configuration, context=context or None, **kwargs)
+            client=kubernetes.config.new_client_from_config(
+                config_file=config_file, client_configuration=client_configuration, context=context or None, **kwargs
+            )
         )
     except MaxRetryError:
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/incluster_config.py
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -118,9 +118,6 @@ def get_client(
         if proxy:
             LOGGER.info(f"Trying to get client using proxy {proxy}")
             client_configuration = client.Configuration()
-            config.load_kube_config(
-                config_file=config_file, client_configuration=client_configuration, persist_config=True
-            )
             client_configuration.proxy = proxy
 
             return kubernetes.dynamic.DynamicClient(
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -121,9 +121,10 @@ def get_client(
                 config_file=config_file, client_configuration=client_configuration, persist_config=True
             )
             client_configuration.proxy = proxy
-            api_client = client.ApiClient(configuration=client_configuration)
 
-            return kubernetes.dynamic.DynamicClient(client=api_client)
+            return kubernetes.dynamic.DynamicClient(
+                client=kubernetes.config.new_client_from_config(config_file=config_file, client_configuration=client_configuration, context=context or None, **kwargs)
+            )
 
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
         LOGGER.info("Trying to get client via new_client_from_config")
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -114,26 +114,17 @@ def get_client(
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
         proxy = proxy or os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+        client_configuration = client.Configuration()
 
         if proxy:
             LOGGER.info(f"Trying to get client using proxy {proxy}")
-            client_configuration = client.Configuration()
             client_configuration.proxy = proxy
-
-            return kubernetes.dynamic.DynamicClient(
-                client=kubernetes.config.new_client_from_config(
-                    config_file=config_file,
-                    client_configuration=client_configuration,
-                    context=context or None,
-                    **kwargs,
-                )
-            )
 
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
         LOGGER.info("Trying to get client via new_client_from_config")
 
         return kubernetes.dynamic.DynamicClient(
-            client=kubernetes.config.new_client_from_config(config_file=config_file, context=context or None, **kwargs)
+            client=kubernetes.config.new_client_from_config(config_file=config_file, client_configuration=client_configuration, context=context or None, **kwargs)
         )
     except MaxRetryError:
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/incluster_config.py
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -115,7 +115,7 @@ def get_client(
         proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
 
         if proxy:
-            LOGGER.info("Trying to get client using proxy %s", proxy)
+            LOGGER.info(f"Trying to get client using proxy {proxy}")
             client_configuration = client.Configuration()
             config.load_kube_config(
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -107,13 +108,26 @@ def get_client(
...
-        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
-        LOGGER.info("Trying to get client via new_client_from_config")
-
         # kubernetes.config.kube_config.load_kube_config sets KUBE_CONFIG_DEFAULT_LOCATION during module import.
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+
+        if proxy:
```
myakove, line 117: `new_client_from_config` get `client_configuration` so proxy can be set there

myakove (changes requested):

ocp_resources/resource.py
sbahar619: proxy can't be set after the client already created, because the client need the proxy to be set before creating the client in the configuration, because it's using the proxy configuration for low level libraries(urllib3) that require it for initialization when the  creating the client.

ocp_resources/resource.py
```diff
@@ -107,13 +108,26 @@ def get_client(
...
         )
     try:
-        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
-        LOGGER.info("Trying to get client via new_client_from_config")
-
         # kubernetes.config.kube_config.load_kube_config sets KUBE_CONFIG_DEFAULT_LOCATION during module import.
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
```
rnetser, line 115: `proxy` should be passed as arg

ocp_resources/resource.py
```diff
@@ -107,13 +108,26 @@ def get_client(
...
+
+        if proxy:
+            LOGGER.info("Trying to get client using proxy %s", proxy)
```
rnetser, line 118: please use f-string formatting

rnetser (changes requested):

ocp_resources/resource.py
sbahar619: Why?
Using those environment variables are a common way to configure proxy in Python...

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
myakove: Seems to me that it should work

```
In [19]: proxy = "https://localost.com"

In [20]: client_configuration = client.Configuration()

In [21]: client_configuration.proxy = proxy

In [22]: qq = kubernetes.dynamic.DynamicClient(client=kubernetes.config.new_client_from_config(config_file=config_file, client_configuration=client_configuration))

In [23]: qq.configuration.proxy
Out[23]: 'https://localost.com'
```

ocp_resources/resource.py
myakove: I don't think this is a common way in Python.

Maybe a common way to define PROXY to applications in general, anyway since oc/kubeconfig use environment variables to get the proxy we can use it here, but you need to add it as arg as well

myakove, line 115: Missing typing.

myakove (changes requested):

ocp_resources/resource.py
sbahar619: Ohh got it now, done, thanks!

ocp_resources/resource.py
```diff
@@ -107,13 +108,32 @@ def get_client(
...
-        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
-        LOGGER.info("Trying to get client via new_client_from_config")
-
         # kubernetes.config.kube_config.load_kube_config sets KUBE_CONFIG_DEFAULT_LOCATION during module import.
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        proxy = kwargs.pop("proxy", None) or os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+
+        if proxy:
```
myakove, line 117: here you just need to prepare `client_configuration` and use the call that we already have in line 137

ocp_resources/resource.py
```diff
@@ -107,13 +108,32 @@ def get_client(
...
         )
     try:
-        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
-        LOGGER.info("Trying to get client via new_client_from_config")
-
         # kubernetes.config.kube_config.load_kube_config sets KUBE_CONFIG_DEFAULT_LOCATION during module import.
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        proxy = kwargs.pop("proxy", None) or os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
```
myakove, line 115: get `proxy` as arg to the function, not as part of kwargs

myakove (changes requested):

ocp_resources/resource.py
sbahar619: Got it, done.
Please note that the additional argument will not take into affect for internal calls of get_client, like in
```
    def _set_client_and_api_version(self) -> None:
        if not self.client:
            self.client = get_client(config_file=self.config_file, context=self.context)

        if not self.api_version:
            self.api_version = _get_api_version(dyn_client=self.client, api_group=self.api_group, kind=self.kind)
```
Which is being called internally by 'exists' property.

sbahar619: Done.

ocp_resources/resource.py
myakove: now that I think of it, we should get `client_configuration` from the user and use it
This way we can support any configuration

ocp_resources/resource.py
```diff
@@ -107,13 +108,34 @@ def get_client(
...
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        proxy: Optional[str] = (
+            kwargs.pop("proxy", None) or os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+        )
+
+        if proxy:
+            LOGGER.info(f"Trying to get client using proxy {proxy}")
+            client_configuration = client.Configuration()
+            config.load_kube_config(
```
myakove, line 122: why this line is needed?

myakove (changes requested):

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Right!
Not required, I removed it.

ocp_resources/resource.py
myakove: .

myakove (changes requested):

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
```diff
@@ -107,15 +113,24 @@ def get_client(
...
     try:
-        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
-        LOGGER.info("Trying to get client via new_client_from_config")
-
         # kubernetes.config.kube_config.load_kube_config sets KUBE_CONFIG_DEFAULT_LOCATION during module import.
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = client_configuration or client.Configuration()
+        proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
```
myakove, line 121: User will give as configured `client_configuration` and we just use it.

myakove (changes requested):

ocp_resources/resource.py
sbahar619: It will make the implementation very complicated on the caller side if we omit this important option.
Please consider keeping the proxy to be set additionally as environment variable.

sbahar619: I can add logic to override the proxy part if it's already being configured in the  client_configuration regardless of being set by environment variable.

sbahar619:
> @sbahar619 this can be closed right? if so please close the PR

Right now yes, but we need to solve this issue, I will write a solution proposal to decide on the agreed solution.

myakove:
> > @sbahar619 this can be closed right? if so please close the PR
> 
> Right now yes, but we need to solve this issue, I will write a solution proposal to decide on the agreed solution.

@sbahar619  Which issue are you trying to solve?

sbahar619:
> > > @sbahar619 this can be closed right? if so please close the PR
> > 
> > 
> > Right now yes, but we need to solve this issue, I will write a solution proposal to decide on the agreed solution.
> 
> @sbahar619 Which issue are you trying to solve?

To configure the client to use proxy, currently to implement it in cnv-tests will be complicated so before I will start working on it, I will create a document about the possible implementation solutions and after agreement I will start.

PR #2186, opened 2024-10-28
Add wait for status condition method

sbahar619:
##### Short description:
Added a wait_for_status_condition function to support flexible, user-defined status checks, allowing for custom success and failure conditions.
##### More details:
The new wait_for_status_condition function introduces a generic approach to checking status conditions, making it adaptable for various scenarios. Unlike [wait_for_condition](https://github.com/RedHatQE/openshift-python-wrapper/blob/c368445ae2319688a0280fbe9f433370e526efa1/ocp_resources/resource.py#L1002), which only handles specific, hardcoded status checks, this function accepts lists of condition-check functions (wait_condition_fns and not_wait_condition_fns) to evaluate both success and failure conditions as needed.

This approach offers:

* Adaptability: Supports any status condition without modifying the function itself.
* Simplified Maintenance: New conditions can be added simply by passing functions, reducing code changes.
* Separation of Concerns: Retains wait_for_condition for straightforward checks, while wait_for_status_condition addresses complex use cases.
## Summary by CodeRabbit

- **New Features**
	- Introduced a new exception for handling status condition failures.
	- Added a method to monitor specific resource status conditions with customizable timeout and interval settings.

- **Bug Fixes**
	- Enhanced error handling for resource status monitoring.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -136,6 +136,11 @@ def sub_resource_level(current_class: Any, owner_class: Any, parent_class: Any) 
 
 # Exceptions classes
 
+
+class WaitForStatusConditionFailed(Exception):
+    """Exception raised when waiting for a status condition fails."""
+
+
 # End Exceptions classes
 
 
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -1028,6 +1033,58 @@ def wait_for_condition(self, condition: str, status: str, timeout: int = 300) ->
...
+        """
+        samples = TimeoutSampler(
+            wait_timeout=wait_timeout, sleep=sleep_interval, func=lambda: self.instance.status.conditions
+        )
+
+        try:
+            for sample in samples:
+                for condition in sample:
+                    if all(wait_fn(condition) for wait_fn in wait_condition_fns):
+                        return condition
```
yossisegev, line 1072: IIUC, this means that this function will return success before checking if an un-wanted condition (i.e. one of the conditions that are covered in the `not_wait_condition_fns` list) is True.

yossisegev (changes requested):

ocp_resources/resource.py
```diff
@@ -136,6 +136,11 @@ def sub_resource_level(current_class: Any, owner_class: Any, parent_class: Any)
 
 # Exceptions classes
 
+
+class WaitForStatusConditionFailed(Exception):
```
dbasunag, line 140: Don't we have a file exceptions.py?

ocp_resources/resource.py
```diff
@@ -1028,6 +1033,58 @@ def wait_for_condition(self, condition: str, status: str, timeout: int = 300) ->
                     if cond["type"] == condition and cond["status"] == status:
                         return
 
+    def wait_for_status_condition(
```
dbasunag, line 1036: I don't understand how this belongs in wrapper. It seems more specific to tests/infra that uses wrapper. e.g. in cnv-tests we have wait_for_consistent_resource_conditions(), if you need more control, may be consider creating a helper there?

dbasunag (changes requested):

ocp_resources/resource.py
sbahar619: Good point, I also reviewed [wait_for_consistent_resource_conditions](https://github.com/RedHatQE/cnv-tests/blob/5611e099b05265688a2c93b922bbedfa3d834c07/utilities/infra.py#L411) and it doing most of the functionalities my PR introduces.
I can extend it there.

ocp_resources/resource.py
sbahar619: Right, why it's a problem?
If the user desired status is met, the unintended status is not relevant.
The user is responsible to give proper condition function without conflicts.

PR #2188, opened 2024-10-30
Add UDN properties and methods

sbahar619:
##### Short description:
Add UDN properties and methods.
## Summary by CodeRabbit

- **New Features**
	- Enhanced `UserDefinedNetwork` class with methods to evaluate network readiness and synchronization status.
	- Introduced properties to retrieve current status conditions, assess readiness, and check for synchronization errors.
  
These updates improve the monitoring and management capabilities of the `UserDefinedNetwork`.

### DIFF
--- a/ocp_resources/user_defined_network.py
+++ b/ocp_resources/user_defined_network.py
@@ -103,7 +103,7 @@ class UserDefinedNetwork(NamespacedResource):
         )
 
     @property
-    def conditions(self) -> List:
+    def conditions(self) -> List[Dict[str, Any]]:
         """
         Retrieve the current status conditions of the UserDefinedNetwork instance.
 
--- a/ocp_resources/user_defined_network.py
+++ b/ocp_resources/user_defined_network.py
@@ -68,6 +68,74 @@ class UserDefinedNetwork(NamespacedResource):
 
     # End of generated code
 
+    @classmethod
+    def is_ready_condition(cls, condition: dict) -> bool:
+        """
+        Check if the given condition indicates that the UserDefinedNetwork is ready.
+
+        Args:
+            condition (dict): A dictionary representing the condition of the UserDefinedNetwork.
+
+        Returns:
+            bool: True if the condition indicates the UserDefinedNetwork is ready, False otherwise.
+        """
+        return (
+            condition.get("reason") == cls.Condition.Reason.NETWORK_ATTACHMENT_DEFINITION_READY
+            and condition.get("status") == cls.Condition.Status.TRUE
+            and condition.get("type") == cls.Condition.Type.NETWORK_READY
+        )
+
+    @classmethod
+    def is_sync_error_condition(cls, condition: dict) -> bool:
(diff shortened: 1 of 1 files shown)


ocp_resources/user_defined_network.py
```diff
@@ -68,6 +68,74 @@ def to_dict(self) -> None:
...
+
+        Args:
+            condition (dict): A dictionary representing the condition of the UserDefinedNetwork.
+
+        Returns:
+            bool: True if the condition indicates a synchronization error, False otherwise.
+        """
+        return (
+            condition.get("reason") == cls.Condition.Reason.SYNC_ERROR
+            and condition.get("status") == cls.Condition.Status.FALSE
```
yossisegev, line 101: If the status is `false`, doesn't it mean that the error condition (SYNC_ERROR) you are checking here is not filled?

yossisegev (changes requested):

ocp_resources/user_defined_network.py
sbahar619: It's not mean the condition is not filled, it means that the condition is not applicable, which is correct as the condition represent sync error, a negative condition.
See k8s documentaion for more info:
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

yossisegev: Oh, my bad. I confused the condition's `reason` with its type. I now see that `SYNC_ERROR` is the `reason`, while the `type` is `NETWORK_READY`, which is false in this check.
Sorry for missing that.

yossisegev (approved):

ocp_resources/user_defined_network.py
```diff
@@ -68,6 +68,74 @@ def to_dict(self) -> None:
...
+            and condition.get("type") == cls.Condition.Type.NETWORK_READY
+        )
+
+    @property
+    def conditions(self) -> List[Dict[str, Any]]:
```
dbasunag, line 106: I don't understand why this is needed. You can already access the conditions using self.instance.status.conditions.

ocp_resources/user_defined_network.py
```diff
@@ -68,6 +68,74 @@ def to_dict(self) -> None:
...
+        Returns:
+            list: A list of status conditions associated with the UserDefinedNetwork instance.
+        """
+        # Cache instance to avoid repeated API requests
+        instance = self.instance
+        if instance.status and instance.status.conditions:
+            return instance.status.conditions
+        return []
+
+    @property
```
dbasunag, line 119: This PR seems better suited for cnv-tests, no?

myakove: Yes!

myakove (changes requested):

ocp_resources/user_defined_network.py
sbahar619: Got it.

PR #2220, opened 2024-12-06
Update Model Registry check on grpc and rest spec

lugi0:
Passing in an empty dict should be allowed, the operator then takes care of filling out the missing bits of the resource.

### DIFF
--- a/ocp_resources/model_registry.py
+++ b/ocp_resources/model_registry.py
@@ -56,10 +56,10 @@ class ModelRegistry(NamespacedResource):
         super().to_dict()
 
         if not self.kind_dict and not self.yaml_file:
-            if not self.grpc:
+            if not self.grpc and self.grpc != {}:
                 raise MissingRequiredArgumentError(argument="self.grpc")
 
-            if not self.rest:
+            if not self.rest and self.rest != {}:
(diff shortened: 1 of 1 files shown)


ocp_resources/model_registry.py
```diff
@@ -56,10 +56,10 @@ def to_dict(self) -> None:
         super().to_dict()
 
         if not self.kind_dict and not self.yaml_file:
-            if not self.grpc:
+            if not self.grpc and self.grpc != {}:
```
myakove, line 59: @rnetser  I think we need to handle this is the class-generator as well.

@lugi0  please do:
```
if self.grpc is None:
```

myakove (changes requested):

ocp_resources/model_registry.py
myakove: https://github.com/RedHatQE/openshift-python-wrapper/pull/2221

myakove: @lugi0 or better, regenerate the file after my PR is merged.

lugi0: Sounds good @myakove

lugi0: Done

PR #2221, opened 2024-12-08
class-generator: check of mandatory arg is  None

myakove:
## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling for required arguments in various classes to raise errors only when explicitly set to `None`.
	- Improved clarity in conditions for `containers` and `control_plane_ref` attributes to ensure correct error triggering.
  
- **Tests**
	- Streamlined the test for comparing generated output files against expected files, improving error handling and clarity.

### DIFF
--- a/class_generator/manifests/class_generator_template.j2
+++ b/class_generator/manifests/class_generator_template.j2
@@ -52,7 +52,7 @@ class {{ kind }}({{ base_class }}):
         if not self.kind_dict and not self.yaml_file:
           {% if all_required_args %}
           {% for arg in all_required_args %}
-            if not self.{{ arg }}:
+            if self.{{ arg }} is None:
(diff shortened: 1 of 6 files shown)


lugi0 (approved):

lugi0 (approved):

lugi0:
For some added context: in #2220 I've update a similar check from `if not self.foo` to `if not self.foo and self.foo != {}`; this is because if `foo` is an empty dictionary in Python it evaluates to falsey in boolean logic while Go treats it as a valid object, thus it should be allowed.
`(not self.foo and self.foo != {}) == (self.foo is None)` are equivalent expressions in my testing, so I'm guessing @myakove made these changes for the same reason.

rnetser (approved):

PR #2240, opened 2024-12-18
Add retry for api_request() call

dbasunag:
##### More details: Recently noticing 500 error while running tests, specially against various `Put` or `Patch` calls:
```
"Internal error occurred: Internal error occurred: failed calling webhook \"mutatevirtualmachines.kubemacpool.io\": failed to call webhook: Post \"https://kubemacpool-service.openshift-cnv.svc:443/mutate-virtualmachines?timeout=10s\
```
These errors recovers and on retry the call succeeds. 

## Release Notes

- **New Features**
	- Enhanced API request handling with configurable retry parameters across multiple resource classes.
	- Added flexible timeout and retry configurations for API interactions.

- **Improvements**
	- Introduced new timeout constants (`TIMEOUT_1SEC`, `TIMEOUT_30SEC`, `TIMEOUT_5SEC`) to support more granular request retry logic.
	- Updated method signatures to provide more control over API request retries.

- **Chores**
	- Added utility constants to support improved error handling and request management.

PR #2250, opened 2024-12-20
replace retry_cluster_exceptions with urllib3.util.Retry

dbasunag:
## Summary by CodeRabbit

- **New Features**
	- Introduced a context manager for enhanced error handling and retry mechanisms in Kubernetes API interactions.
- **Bug Fixes**
	- Improved error handling in API calls, ensuring consistent retry behavior across resource management functions.
- **Refactor**
	- Centralized retry logic, simplifying method signatures and enhancing maintainability.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -14,13 +14,14 @@ from io import StringIO
 from signal import SIGINT, signal
 from types import TracebackType
 from typing import Optional, Any, Dict, List
+import urllib3
+import http
 
 import kubernetes
 from kubernetes.dynamic import DynamicClient, ResourceInstance
 import yaml
 from benedict import benedict
 from kubernetes.dynamic.exceptions import (
-    ConflictError,
     MethodNotAllowedError,
     NotFoundError,
     ForbiddenError,
@@ -37,9 +38,7 @@ from ocp_resources.constants import (
     PROTOCOL_ERROR_EXCEPTION_DICT,
     TIMEOUT_1MINUTE,
     TIMEOUT_4MINUTES,
-    TIMEOUT_10SEC,
     TIMEOUT_30SEC,
-    TIMEOUT_5SEC,
 )
 from ocp_resources.event import Event
 from timeout_sampler import (
@@ -50,6 +49,14 @@ from timeout_sampler import (
 from ocp_resources.exceptions import MissingRequiredArgumentError, MissingResourceResError
 from ocp_resources.utils import skip_existing_resource_creation_teardown
 
+DEFAULT_HTTP_STATUS_CODE_TO_RETRY = urllib3.util.Retry.RETRY_AFTER_STATUS_CODES | {
+    http.HTTPStatus.INTERNAL_SERVER_ERROR.value,
+    http.HTTPStatus.GATEWAY_TIMEOUT.value,
+    http.HTTPStatus.FORBIDDEN.value,
+    http.HTTPStatus.CONFLICT.value,
+    http.HTTPStatus.NOT_FOUND.value,
+    104,
+}
 
 LOGGER = get_logger(name=__name__)
 MAX_SUPPORTED_API_VERSION = "v2"
@@ -74,6 +81,29 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
 
     LOGGER.info(f"kind: {kind} api version: {res.group_version}")
     return res.group_version
+
+
+@contextlib.contextmanager
+def get_retry_cluster_exceptions_client(
+    client: DynamicClient,
+    backoff_factor: float = 0.5,
+    retry_count: int = 5,
+    status_codes=None,
+) -> Any:
+    original_retries = client.configuration.retries
+    if not status_codes:
+        status_codes = set()
+    client.configuration.retries = urllib3.util.Retry(
+        total=retry_count,
+        backoff_factor=backoff_factor,
+        allowed_methods=(urllib3.util.Retry.DEFAULT_ALLOWED_METHODS | {http.HTTPMethod.PATCH, http.HTTPMethod.POST}),
+        status_forcelist=(DEFAULT_HTTP_STATUS_CODE_TO_RETRY | status_codes),
+        raise_on_redirect=False,
+        raise_on_status=False,
+    )
+
+    yield client
+    client.configuration.retries = original_retries
 
 
 def get_client(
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -76,6 +83,29 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
     return res.group_version
 
 
+@contextlib.contextmanager
+def get_retry_cluster_exceptions_client(
+    client: DynamicClient,
+    backoff_factor: float = 0.5,
+    retry_count: int = 5,
+    status_codes=None,
```
rnetser, line 91:
```suggestion
    status_codes: set | None = None,
```

please add docstring

ocp_resources/resource.py
```diff
@@ -76,6 +83,29 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
...
+) -> Any:
+    original_retries = client.configuration.retries
+    if not status_codes:
+        status_codes = set()
```
rnetser, line 95:
```suggestion
    status_codes = status_codes or set()
```

ocp_resources/resource.py
```diff
@@ -76,6 +83,29 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
...
+    client.configuration.retries = urllib3.util.Retry(
+        total=retry_count,
+        backoff_factor=backoff_factor,
+        allowed_methods=(urllib3.util.Retry.DEFAULT_ALLOWED_METHODS | {http.HTTPMethod.PATCH, http.HTTPMethod.POST}),
+        status_forcelist=(DEFAULT_HTTP_STATUS_CODE_TO_RETRY | status_codes),
```
rnetser, line 100: the fact that `DEFAULT_HTTP_STATUS_CODE_TO_RETRY` is added by default is hidden here, please add to dosctring

@myakove do we want this behavior? i.e to force apply the defaults?

ocp_resources/resource.py
```diff
@@ -964,41 +994,15 @@ def update_replace(self, resource_dict: Dict[str, Any]) -> None:
...
-            )
-            for sample in sampler:
-                return sample
-
-        except TimeoutExpiredError as exp:
-            if exp.last_exp:
-                raise exp.last_exp
-
-            raise
+        with get_retry_cluster_exceptions_client(client=self.client, retry_count=5, backoff_factor=1) as self.client:
```
rnetser, line 997: `self.api` is a property so we cannot pass it a client but not sure using `as self.client` is correct, @myakove wdyt?

ocp_resources/resource.py
```diff
@@ -1046,11 +1046,11 @@ def instance(self) -> ResourceInstance:
...
         """
+        _instance_out = None
 
-        def _instance() -> Optional[ResourceInstance]:
-            return self.api.get(name=self.name)
-
-        return self.retry_cluster_exceptions(func=_instance)
+        with get_retry_cluster_exceptions_client(client=self.client) as self.client:
+            _instance_out = self.api.get(name=self.name)
+        return _instance_out
```
rnetser, line 1053:
```
        with get_retry_cluster_exceptions_client(client=self.client) as self.client:
           return self.api.get(name=self.name)
```

ocp_resources/resource.py
```diff
@@ -1131,17 +1131,18 @@ def api_request(self, method: str, action: str, url: str, **params: Any) -> Dict
...
+            )
 
-        try:
-            return json.loads(response.data)
-        except json.decoder.JSONDecodeError:
-            return response.data
+            try:
+                return json.loads(response.data)
+            except json.decoder.JSONDecodeError:
+                return response.data
```
rnetser, line 1145: why does this block need to be under the `with`?

ocp_resources/resource.py
```diff
@@ -1402,11 +1399,11 @@ def instance(self) -> ResourceInstance:
...
         """
+        _instance_out = None
 
-        def _instance() -> ResourceInstance:
-            return self.api.get(name=self.name, namespace=self.namespace)
-
-        return self.retry_cluster_exceptions(func=_instance)
+        with get_retry_cluster_exceptions_client(client=self.client) as self.client:
+            _instance_out = self.api.get(name=self.name, namespace=self.namespace)
+        return _instance_out
```
rnetser, line 1406:
```suggestion
        with get_retry_cluster_exceptions_client(client=self.client) as self.client:
           return self.api.get(name=self.name, namespace=self.namespace)

```

rnetser (changes requested):

ocp_resources/resource.py
```diff
@@ -50,6 +49,14 @@
 from ocp_resources.exceptions import MissingRequiredArgumentError, MissingResourceResError
 from ocp_resources.utils import skip_existing_resource_creation_teardown
 
+DEFAULT_HTTP_STATUS_CODE_TO_RETRY = urllib3.util.Retry.RETRY_AFTER_STATUS_CODES | {
+    http.HTTPStatus.INTERNAL_SERVER_ERROR.value,
+    http.HTTPStatus.GATEWAY_TIMEOUT.value,
+    http.HTTPStatus.FORBIDDEN.value,
+    http.HTTPStatus.CONFLICT.value,
```
dollierp, line 56: Usually after a  `HTTPStatus.CONFLICT`, urllib3 retry mechanism won't help. You need to update the `metadata.resourceVersion` field before retrying.

ocp_resources/resource.py
```diff
@@ -76,6 +83,29 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
...
+    original_retries = client.configuration.retries
+    if not status_codes:
+        status_codes = set()
+    client.configuration.retries = urllib3.util.Retry(
+        total=retry_count,
+        backoff_factor=backoff_factor,
+        allowed_methods=(urllib3.util.Retry.DEFAULT_ALLOWED_METHODS | {http.HTTPMethod.PATCH, http.HTTPMethod.POST}),
+        status_forcelist=(DEFAULT_HTTP_STATUS_CODE_TO_RETRY | status_codes),
+        raise_on_redirect=False,
+        raise_on_status=False,
```
dollierp, line 102: Note that it might change the exposed "API" of the openshift-python-wrapper. It will not raise MaxRetryError anymore, instead it will raise the underlying exception (usually a kubernetes.ApiException). IMHO, this is a better behaviour but it might not be backward compatible.

ocp_resources/resource.py
```diff
@@ -76,6 +83,29 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
...
+def get_retry_cluster_exceptions_client(
+    client: DynamicClient,
+    backoff_factor: float = 0.5,
+    retry_count: int = 5,
+    status_codes=None,
+) -> Any:
+    original_retries = client.configuration.retries
+    if not status_codes:
+        status_codes = set()
+    client.configuration.retries = urllib3.util.Retry(
```
dollierp, line 96: I think modifying the `configuration.retries` attribute after the client has been created has no effect.

AFAIK, you need to first create the configuration object, set its retry attribute and create the client from the configuration object.

ocp_resources/resource.py
dbasunag: @myakove / @rnetser please let me know how to proceed here.

ocp_resources/resource.py
dbasunag: We get ConflictError during ResourceEditor calls often. Using timeout sampler this currently works. If urllib3 retry won't work the same way, then we have a problem here. @myakove I would need your suggestion here.

ocp_resources/resource.py
dbasunag: I do currently see the client is updated with the retry attributes. Are you saying that changes won't help? In that case wrapper would not be the right place for these changes.

ocp_resources/resource.py
rnetser: i'd add an option to use defaults

dbasunag:
Closing this one in favor of expanding wrapper's existing retry logic with https://github.com/RedHatQE/openshift-python-wrapper/pull/2240, based on my chat with @myakove. Wrapper's retry is more flexible than what we can do with urllib3.utils.Retry.

PR #2289, opened 2025-01-22
Update wait_for_status_success() call to look at both type and status for status.conditions

dbasunag:
## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced status checking logic for Node Network Configuration Policy
  - Improved validation of configuration status conditions
  - Streamlined control flow and error handling in the status checking process
  - Increased sleep interval for status sampling from 1 second to 5 seconds

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ dev-dependencies = [ "ipdb>=0.13.13", "ipython>=8.12.3" ]
 [project]
 requires-python = ">=3.9"
 name = "openshift-python-wrapper"
-version = "11.0.20"
+version = "11.0.21"
 description = "Wrapper around https://github.com/kubernetes-client/python"
 readme = "README.md"
 license = "Apache-2.0"
--- a/uv.lock
+++ b/uv.lock
@@ -691,7 +691,7 @@ wheels = [
 
 [[package]]
 name = "openshift-python-wrapper"
-version = "11.0.20"
+version = "11.0.21"
 source = { editable = "." }
 dependencies = [
     { name = "click" },


ocp_resources/node_network_configuration_policy.py
```diff
@@ -329,7 +329,10 @@ def _absent_interface(self):
     @property
     def status(self):
         for condition in self.instance.status.conditions:
-            if condition["type"] == self.Conditions.Type.AVAILABLE:
+            if (
```
myakove, line 332: This function returns the status as is.
this is not wait_for_status function

dbasunag: it is not returning the whole condition, just the condition["reason"]. And is only used in ```ocp_resources/node_network_configuration_policy.py:        samples = TimeoutSampler(wait_timeout=self.success_timeout, sleep=1, func=lambda: self.status)
```

File: ocp_resources/node_network_configuration_policy.py [hunk 2]
```diff
@@ -370,32 +370,26 @@ def _process_failed_status(self, failed_condition_reason):
...
-                f"{f'nodes: {[node.name for node in self.nodes]}' if self.nodes else ''}"
-            )
-            raise
+        available_condition = [
+            condition
+            for condition in self.instance.get("status", {}).get("conditions", [])
+            if condition and condition["type"] == self.Conditions.Type.AVAILABLE
+        ]
+        if available_condition:
+            if available_condition[0]["status"] == self.Condition.Status.TRUE:
```
myakove, line 384: `available_condition` is a list, so I guess it can have more than one? 
If so please make sure you operate on the right one and not use [0] by default.

And i this is the case the name should be `available_conditions`

ocp_resources/node_network_configuration_policy.py
```diff
@@ -370,32 +370,26 @@ def _process_failed_status(self, failed_condition_reason):
 
         raise NNCPConfigurationFailed(f"Reason: {failed_condition_reason}\n{last_err_msg}")
 
+    @retry(wait_timeout=120, sleep=5)
```
myakove, line 373: `120` is problematic, you drop the ability to get the `wait_timeout`  (`self.success_timeout`) from the user.
I don't think `@retry` fit here

ocp_resources/node_network_configuration_policy.py
```diff
@@ -370,32 +370,26 @@ def _process_failed_status(self, failed_condition_reason):
...
     def wait_for_status_success(self):
         failed_condition_reason = self.Conditions.Reason.FAILED_TO_CONFIGURE
         no_match_node_condition_reason = self.Conditions.Reason.NO_MATCHING_NODE
 
-        # if we get here too fast there are no conditions, we need to wait.
-        self.wait_for_configuration_conditions_unknown_or_progressing()
```
myakove, line 378: This is not needed anymore?

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
dbasunag: Does not look like. I ran the tests many times.

ocp_resources/node_network_configuration_policy.py
dbasunag: No, in status.conditions, there is only one type that is "Available". So it would only be one value in this list - if we find a match.

ocp_resources/node_network_configuration_policy.py
myakove: then I think you can remove the function as well.

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
dbasunag: I checked. It is currently used in https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/network/nmstate/test_nmstate_sanity.py#L211

myakove: ok

ocp_resources/node_network_configuration_policy.py
```diff
@@ -374,20 +374,25 @@ def wait_for_status_success(self):
...
+            return conditions[0] if conditions else {}
 
-        samples = TimeoutSampler(wait_timeout=self.success_timeout, sleep=1, func=lambda: self.status)
         try:
-            for sample in samples:
-                if sample == self.Conditions.Reason.SUCCESSFULLY_CONFIGURED:
+            for sample in TimeoutSampler(
+                wait_timeout=self.success_timeout, sleep=5, func=_get_available_status_condition
+            ):
+                if sample.get("status") == self.Condition.Status.TRUE:
```
myakove, line 389: Maybe it's better to do here
```
if not sample:
    continue
```

dbasunag: Since I am returning {} if no condition is found, I don't think I need the check for sample. if sample.get("status") would be None by default.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -374,20 +374,25 @@ def wait_for_status_success(self):
         failed_condition_reason = self.Conditions.Reason.FAILED_TO_CONFIGURE
         no_match_node_condition_reason = self.Conditions.Reason.NO_MATCHING_NODE
 
-        # if we get here too fast there are no conditions, we need to wait.
-        self.wait_for_configuration_conditions_unknown_or_progressing()
+        def _get_available_status_condition() -> dict[str, str]:
```
EdDev, line 377: I do not like these closures (which are not closures).
How about this:
```python
class NodeNetworkConfigurationPolicy(Resource):
  def wait_for_status_success(self):
    ...
    try:
      return _wait_for_available_condition()
    except (TimeoutExpiredError, NNCPConfigurationFailed):
      ...
      raise
  
  @retry(wait_timeout=480, sleep=5, exceptions_dict={SomeNoopError: []},)
  def _wait_for_available_condition(self):
    available_conditions = (
                condition
                for condition in self.instance.get("status", {}).get("conditions", [])
                if condition and condition["type"] == self.Conditions.Type.AVAILABLE
    )
    for condition in available_conditions:
      if condition.get("status") == self.Condition.Status.TRUE:
        ...
      elif condition.get("reason") == no_match_node_condition_reason:
        ...
      elif sample.get("reason") == failed_condition_reason:
        ...
      return None
    return None
```

The only disadvantage is that you cannot use `self.success_timeout` in the decorator.
But I think you can use the upper value to exit earlier.
You also need the `SomeNoopError` so it will not retry in case there is a error (no matter which one).
If there are errors you do want to retry over, you should add it instead of it.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -370,26 +370,37 @@ def _process_failed_status(self, failed_condition_reason):
...
+            conditions = [
+                condition
+                for condition in self.instance.get("status", {}).get("conditions", [])
+                if condition and condition["type"] == self.Conditions.Type.AVAILABLE
+            ]
+            return conditions[0] if conditions else {}
+
+        try:
+            for sample in TimeoutSampler(
+                wait_timeout=self.success_timeout, sleep=5, func=_get_available_status_condition
```
EdDev, line 387: If you cannot use the `retry`, then maybe remove that inner function with:
```python
`func=lambda: next((
            condition
            for condition in self.instance.get("status", {}).get("conditions", [])
            if condition and condition["type"] == self.Conditions.Type.AVAILABLE
), {})
```

EdDev (approved):
The logic looks right to me, thanks!

I do not like some parts, but it is not worth blocking on them.

ocp_resources/node_network_configuration_policy.py
dbasunag: retry would require hard coding wait_time. But I see that won't be a good idea even for cnv. Since we use success_timeout ourselves.

ocp_resources/node_network_configuration_policy.py
dbasunag: Yes, this would be cleaner. Let me try it and see how that goes.

EdDev (approved):

ocp_resources/node_network_configuration_policy.py
myakove: Since you can tell that sample may be empty you can skip all other calls.
There is no point to check 3 different things while you know that sample is empty.

Yes, the code will not fail, but it's not optimized.

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
```diff
@@ -374,21 +374,28 @@ def wait_for_status_success(self):
...
-
-                elif sample == no_match_node_condition_reason:
-                    raise NNCPConfigurationFailed(f"{self.name}. Reason: {no_match_node_condition_reason}")
-
-                elif sample == failed_condition_reason:
-                    self._process_failed_status(failed_condition_reason=failed_condition_reason)
+            for sample in TimeoutSampler(
+                wait_timeout=self.success_timeout,
+                sleep=5,
+                func=lambda: next(
```
myakove, line 381: Why did you change it? It was more clean and readable before.

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
dbasunag: Based on https://github.com/RedHatQE/openshift-python-wrapper/pull/2289#discussion_r1927430258 I changed it. Please let me know what you want me to do next.

EdDev: @myakove , I thought you care about efficiency. This is more efficient than re-defining the function on each call, and using generators to reduce memory consumption. And regarding readability, it’s opinionated, for me this is shorter and clearer.
However, this is your call, none of the reasons above merit blocking the PR.

myakove: ack

myakove (approved):

ocp_resources/node_network_configuration_policy.py
```diff
@@ -374,21 +374,28 @@ def wait_for_status_success(self):
...
+            ):
+                if sample:
+                    if sample["status"] == self.Condition.Status.TRUE:
+                        self.logger.info(f"NNCP {self.name} configured Successfully")
+                        return sample
+                    elif sample["reason"] == no_match_node_condition_reason:
+                        raise NNCPConfigurationFailed(f"{self.name}. Reason: {no_match_node_condition_reason}")
+
+                    elif sample["reason"] == failed_condition_reason:
+                        self._process_failed_status(failed_condition_reason=failed_condition_reason)
```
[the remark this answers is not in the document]
myakove: @dbasunag @EdDev Is this valid comment?
once we have `sample` we will always have the keys there or we need to check them?

rnetser (dismissed):

ocp_resources/node_network_configuration_policy.py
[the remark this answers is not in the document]
EdDev: > once we have sample we will always have the keys there or we need to check them?

Only the `type` and `status` are a must, the rest are optional.
I would use the `sample.get` version here.
Nice catch.

Ref: https://github.com/nmstate/kubernetes-nmstate/blob/main/api/shared/condition_types.go#L30-L31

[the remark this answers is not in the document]
dbasunag: Will change the reason.

hmeir (approved):

rnetser (approved):

myakove (approved):

redhat-qe-bot2:
Cherry-picked PR Update wait_for_status_success() call to look at both type and status for status.conditions into v4.16

Issue #2182, opened 2024-10-27
class-generator: fix when kind do not have `properties` - 2181


redhat-qe-bot1:
[Auto generated]
Number: [#2181]

PR #2291, opened 2025-01-26
Add proxy support to Kubernetes client configuration in get_client

sbahar619:
##### Short description:
Add proxy support to Kubernetes client configuration in get_client
##### More details:
This update modifies the get_client function to support setting a proxy for the Kubernetes client configuration. The client_configuration argument is introduced to allow passing a client.Configuration object directly, and if not provided, it defaults to creating a new configuration instance. Additionally, if the HTTPS_PROXY or HTTP_PROXY environment variables are set, the proxy is applied to the Kubernetes client configuration.

Introduced dynamic proxy enablement by reading the OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY environment variable.
Updated README.md with instructions on using the new proxy enablement feature.
This solution is preferred over requiring the caller to set the client.Configuration for each client instance creation. Doing so would necessitate updating every instance where the client is created, which would be less efficient and error-prone.
https://issues.redhat.com/browse/CNV-46351
A similar solution was previously proposed in PR #2066 , but it was not accepted at the time. This PR aims to provide additional reasoning and demonstrate why this approach is the most effective.
This change is especially useful in environments where proxies are required, and it simplifies the client setup process while minimizing the chances of errors.

Note:
We can also promote this upstream PR in kubernetes-client, which is a good solution:
https://github.com/kubernetes-client/python/pull/2182
## Summary by CodeRabbit

- **New Features**
	- Enhanced Kubernetes client configuration with optional proxy support.
	- Added ability to dynamically enable proxy settings through environment variables.

- **Bug Fixes**
	- Implemented error handling for conflicting proxy settings in client configuration.

- **Documentation**
	- Introduced a new section in the README for "Proxy Enablement" with instructions for configuring proxy settings.

- **Tests**
	- Added tests to verify error handling for conflicting proxy settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,11 +97,16 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
 
 To enable proxy configuration for the client:
 
-Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>
-This enables the proxy without requiring the use_proxy argument when creating the client.
+1. Set the `use_proxy` argument when creating the client or set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
+This enables the proxy without requiring the `use_proxy` argument when creating the client.
 
-Define the HTTPS_PROXY or HTTP_PROXY environment variable for the proxy URL.
-If neither is set, a ValueError will be raised.
+2. Define either `HTTPS_PROXY` or `HTTP_PROXY` environment variable with your proxy URL:
+```bash
+export HTTPS_PROXY="http://proxy.example.com:8080"
+# or
+export HTTP_PROXY="http://proxy.example.com:8080"
+```
+If neither variable is set when proxy is enabled, a `ValueError` will be raised.
 
 ## Code check
 
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
 
 To enable proxy configuration for the client:
 
-1. Set the `use_proxy` argument when creating the client or set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
+1. Set the `use_proxy` argument to `True` when creating the client or set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
 This enables the proxy without requiring the `use_proxy` argument when creating the client.
 
 2. Define either `HTTPS_PROXY` or `HTTP_PROXY` environment variable with your proxy URL:
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
 
 To dynamically enable proxy configuration:
 
-Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>.
+Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>
 This will automatically enable the proxy without needing to set the use_proxy argument when creating the client.
 
 Make sure to define the HTTPS_PROXY or HTTP_PROXY environment variable for the proxy URL.
--- a/README.md
+++ b/README.md
@@ -93,6 +93,16 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
   export OPENSHIFT_PYTHON_WRAPPER_HASH_LOG_DATA="false"
   ```
 
+## Dynamic Proxy Enablement
+
+To dynamically enable proxy configuration:
+
+Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=1.
+This will automatically enable the proxy without needing to set the use_proxy argument when creating the client.
+
+Make sure to define the HTTPS_PROXY or HTTP_PROXY environment variable for the proxy URL.
+If neither is set, a ValueError will be raised.
+
 ## Code check
 
 We use pre-commit for code check.
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
 
 To dynamically enable proxy configuration:
 
-Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=1.
+Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>.
 This will automatically enable the proxy without needing to set the use_proxy argument when creating the client.
 
 Make sure to define the HTTPS_PROXY or HTTP_PROXY environment variable for the proxy URL.
--- a/README.md
+++ b/README.md
@@ -96,12 +96,11 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
 ## Proxy Enablement
 
 This configuration allows the client to route traffic through a specified proxy server.
-It can be enabled via the `use_proxy` argument or environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY`.
+It can be enabled via the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY`.
 
 To enable proxy configuration for the client:
 
-1. Set the `use_proxy` argument to `True` when creating the client or set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
-This enables the proxy without requiring the `use_proxy` argument when creating the client.
+1. Set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
 
 2. Define either `HTTPS_PROXY` or `HTTP_PROXY` environment variable with your proxy URL:
 ```bash
--- a/README.md
+++ b/README.md
@@ -93,14 +93,14 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
   export OPENSHIFT_PYTHON_WRAPPER_HASH_LOG_DATA="false"
   ```
 
-## Dynamic Proxy Enablement
+## Proxy Enablement
 
-To dynamically enable proxy configuration:
+To enable proxy configuration for the client:
 
 Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>
-This will automatically enable the proxy without needing to set the use_proxy argument when creating the client.
+This enables the proxy without requiring the use_proxy argument when creating the client.
 
-Make sure to define the HTTPS_PROXY or HTTP_PROXY environment variable for the proxy URL.
+Define the HTTPS_PROXY or HTTP_PROXY environment variable for the proxy URL.
 If neither is set, a ValueError will be raised.
 
 ## Code check
--- a/README.md
+++ b/README.md
@@ -95,6 +95,9 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
 
 ## Proxy Enablement
 
+This configuration allows the client to route traffic through a specified proxy server.
+It can be enabled via the `use_proxy` argument or environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY`.
+
 To enable proxy configuration for the client:
 
 1. Set the `use_proxy` argument to `True` when creating the client or set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -100,6 +100,7 @@ def get_client(
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
         use_proxy (bool): If True, retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
+             If the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` is set, it will override the value of `use_proxy`.
         context (str): name of the context to use.
 
     Returns:
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -124,7 +124,7 @@ def get_client(
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
         client_configuration = client_configuration or kubernetes.client.Configuration()
 
-        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
             use_proxy = True
 
         if use_proxy:
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -129,7 +129,6 @@ def get_client(
                 raise ValueError(
                     "Proxy configuration requested but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set"
                 )
-            LOGGER.info(f"Setting proxy in client configuration: {proxy}")
             client_configuration = kubernetes.client.Configuration()
             client_configuration.proxy = proxy
 
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -121,7 +121,7 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
-        client_configuration = kwargs.get('client_configuration', None)
+        client_configuration = kwargs.get("client_configuration", None)
 
         if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
             proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -136,7 +136,7 @@ def get_client(
         return kubernetes.dynamic.DynamicClient(
             client=kubernetes.config.new_client_from_config(
                 config_file=config_file,
-                client_configuration=client_configuration if client_configuration else None,
+                client_configuration=client_configuration,
                 context=context or None,
                 **kwargs,
             )
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -101,8 +101,7 @@ def get_client(
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
         client_configuration (kubernetes.client.Configuration): The kubernetes.client.Configuration to set configs to.
-        use_proxy (bool): If True, automatically retrieves and sets the proxy from the environment
-              variables (`HTTPS_PROXY` or `HTTP_PROXY`). If proxy is not found, raises an error.
+        use_proxy (bool): If True, retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
         context (str): name of the context to use.
 
     Returns:
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -79,7 +79,12 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
 
 
 def get_client(
-    config_file: str = "", config_dict: dict[str, Any] | None = None, context: str = "", **kwargs: Any
+    config_file: str = "",
+    config_dict: dict[str, Any] | None = None,
+    client_configuration: kubernetes.client.Configuration | None = None,
+    use_proxy: bool = False,
+    context: str = "",
+    **kwargs: Any,
 ) -> DynamicClient:
     """
     Get a kubernetes client.
@@ -95,6 +100,10 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        client_configuration (kubernetes.client.Configuration): The kubernetes.client.Configuration to set configs to.
+        use_proxy (bool): If True, automatically retrieves and sets the proxy from the environment
+              variables (`HTTPS_PROXY` or `HTTP_PROXY`). If proxy is not found, raises an error.
+              Default is False, meaning no proxy is used.
         context (str): name of the context to use.
 
     Returns:
@@ -115,8 +124,22 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = client_configuration or kubernetes.client.Configuration()
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") == "1":
+            use_proxy = True
+
+        if use_proxy:
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError("Proxy not found. Please set the HTTPS_PROXY or HTTP_PROXY environment variable.")
+            LOGGER.info(f"Setting proxy in client configuration: {proxy}")
+            client_configuration.proxy = proxy
+
         return kubernetes.dynamic.DynamicClient(
-            client=kubernetes.config.new_client_from_config(config_file=config_file, context=context or None, **kwargs)
+            client=kubernetes.config.new_client_from_config(
+                config_file=config_file, client_configuration=client_configuration, context=context or None, **kwargs
+            )
         )
     except MaxRetryError:
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/incluster_config.py
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -123,10 +123,7 @@ def get_client(
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
         client_configuration = None
 
-        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
-            use_proxy = True
-
-        if use_proxy:
+        if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
             proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
             if not proxy:
                 raise ValueError("Proxy not found. Please set the HTTPS_PROXY or HTTP_PROXY environment variable.")
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -81,7 +81,7 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
 def get_client(
     config_file: str = "",
     config_dict: dict[str, Any] | None = None,
-    use_proxy: bool = False,
+    use_proxy: bool | None = None,
     context: str = "",
     **kwargs: Any,
 ) -> DynamicClient:
@@ -99,9 +99,11 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
-        use_proxy (bool): If True, retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
-            The `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` environment variable, if set and not empty, will be used
-            in addition to the `use_proxy` parameter.
+        use_proxy (bool, default=None):
+            - `True`: retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
+            - `False`: Disables proxy.
+            - `None`: (default): Proxy is determined by the `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` environment
+                  variable. non-empty value enables proxy.
         context (str): name of the context to use.
 
     Returns:
@@ -124,7 +126,7 @@ def get_client(
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
 
-        if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+        if use_proxy or (use_proxy is None and os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY")):
             proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
             if not proxy:
                 raise ValueError(
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -126,7 +126,9 @@ def get_client(
         if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
             proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
             if not proxy:
-                raise ValueError("Proxy not found. Please set the HTTPS_PROXY or HTTP_PROXY environment variable.")
+                raise ValueError(
+                    "Proxy configuration requested but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set"
+                )
             LOGGER.info(f"Setting proxy in client configuration: {proxy}")
             client_configuration = kubernetes.client.Configuration()
             client_configuration.proxy = proxy
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -100,8 +100,8 @@ def get_client(
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
         use_proxy (bool): If True, retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
-            If `use_proxy` is not set, the `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` environment variable will be used
-            if it is set and not empty.
+            The `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` environment variable, if set and not empty, will be used
+            in addition to the `use_proxy` parameter.
         context (str): name of the context to use.
 
     Returns:
@@ -114,7 +114,7 @@ def get_client(
                 config_dict=config_dict, context=context or None, **kwargs
             )
         )
-    client_configuration = kwargs.get("client_configuration")
+    client_configuration = kwargs.get("client_configuration", kubernetes.client.Configuration())
     try:
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
         LOGGER.info("Trying to get client via new_client_from_config")
@@ -128,10 +128,9 @@ def get_client(
             proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
             if not proxy:
                 raise ValueError(
-                    "Proxy configuration requested but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set"
+                    "Proxy configuration is enabled but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set."
                 )
-            if not client_configuration:
-                client_configuration = kubernetes.client.Configuration()
+            if not kwargs.get("client_configuration"):
                 client_configuration.proxy = proxy
 
         return kubernetes.dynamic.DynamicClient(
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -103,7 +103,6 @@ def get_client(
         client_configuration (kubernetes.client.Configuration): The kubernetes.client.Configuration to set configs to.
         use_proxy (bool): If True, automatically retrieves and sets the proxy from the environment
               variables (`HTTPS_PROXY` or `HTTP_PROXY`). If proxy is not found, raises an error.
-              Default is False, meaning no proxy is used.
         context (str): name of the context to use.
 
     Returns:
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -123,8 +123,7 @@ def get_client(
                 raise ValueError(
                     "Proxy configuration is enabled but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set."
                 )
-            if not kwargs.get("client_configuration"):
-                client_configuration.proxy = proxy
+            client_configuration.proxy = proxy
 
         return kubernetes.dynamic.DynamicClient(
             client=kubernetes.config.new_client_from_config(
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -78,7 +78,6 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
 def get_client(
     config_file: str = "",
     config_dict: dict[str, Any] | None = None,
-    use_proxy: bool | None = None,
     context: str = "",
     **kwargs: Any,
 ) -> DynamicClient:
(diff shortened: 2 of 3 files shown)


redhat-qe-bot1:
Report bugs in [Issues](https://github.com/myakove/github-webhook-server/issues)

The following are automatically added:
 * Add reviewers from OWNER file (in the root of the repository) under reviewers section.
 * Set PR size label.
 * New issue is created for the PR. (Closed when PR is merged/closed)
 * Run [pre-commit](https://pre-commit.ci/) if `.pre-commit-config.yaml` exists in the repo.

Available user actions:
 * To mark PR as WIP comment `/wip` to the PR, To remove it from the PR comment `/wip cancel` to the PR.
 * To block merging of PR comment `/hold`, To un-block merging of PR comment `/hold cancel`.
 * To mark PR as verified comment `/verified` to the PR, to un-verify comment `/verified cancel` to the PR.
        verified label removed on each new commit push.
 * To cherry pick a merged PR comment `/cherry-pick <target branch to cherry-pick to>` in the PR.
    * Multiple target branches can be cherry-picked, separated by spaces. (`/cherry-pick branch1 branch2`)
    * Cherry-pick will be started when PR is merged
 * To build and push container image command `/build-and-push-container` in the PR (tag will be the PR number).
    * You can add extra args to the Podman build command
        * Example: `/build-and-push-container --build-arg OPENSHIFT_PYTHON_WRAPPER_COMMIT=<commit_hash>`
 * To add a label by comment use `/<label name>`, to remove, use `/<label name> cancel`
 * To assign reviewers based on OWNERS file use `/assign-reviewers`
 * To check if PR can be merged use `/check-can-merge`
 * to assign reviewer to PR use `/assign-reviewer @<reviewer>`

<details>
<summary>Supported /retest check runs</summary>

 * `/retest tox`: Retest tox
 * `/retest python-module-install`: Retest python-module-install
 * `/retest all`: Retest all

</details>

<details>
<summary>Supported labels</summary>

 * hold
 * verified
 * wip
 * lgtm

</details>

ocp_resources/resource.py
```diff
@@ -2,6 +2,7 @@
 
 from collections.abc import Callable, Generator
 import contextlib
+from kubernetes import client
```
myakove, line 5: `kubernetes` is already imported.

ocp_resources/resource.py
```diff
@@ -115,8 +121,17 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = client_configuration or client.Configuration()
+        proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+
+        if proxy:
+            LOGGER.info(f"Trying to get client using proxy {proxy}")
```
myakove, line 128: The log is not right, you set proxy to the configuration not trying to get client here.

ocp_resources/resource.py
```diff
@@ -115,8 +121,17 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = client_configuration or client.Configuration()
+        proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
```
myakove, line 125: `HTTPS_PROXY` and `HTTP_PROXY` are global proxy config that use by many applications.
I don't think we should set the proxy just by check this, we need the user to request to use proxy and only then try to load those (and fail if not exists)

myakove (changes requested):

ocp_resources/resource.py
sbahar619: ack

ocp_resources/resource.py
sbahar619: ack

sbahar619: Done.

ocp_resources/resource.py
sbahar619: Done.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-python-wrapper/compare/043a0f79d13a1596104e0ad2fd286bb9ff310362..e6d27e64be31f58038a60c956307f4d51025dfd4
REsolve @myakove comments.

ocp_resources/resource.py
sbahar619: I see your point about only using the proxy when explicitly requested by the user. Instead of relying on the generic environment variables, would you prefer defining a project-specific environment variable, such as OPENSHIFT_PYTHON_WRAPPER_USE_PROXY? This would make the proxy setting more explicit and ensure it’s clear that it’s specific to this project.

myakove: No, I mean get `use_proxy` bool arg (default False) and if True call
`proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")`
if `proxy` is None raise error

myakove (changes requested):

ocp_resources/resource.py
sbahar619: The purpose of this PR is to automate the proxy configuration process, so the caller doesn't have to make any changes when creating the client instance. By using the environment variables for the proxy configuration, we can seamlessly integrate the proxy setup into the client, avoiding the need for each caller to manually manage it. This keeps things simple for the user and reduces the chances of errors.

Introducing the use_proxy argument would require the user to set it explicitly, which goes against the goal of this PR. The intention is to automatically handle the proxy setup without adding extra complexity or requiring changes to the existing caller code.

This is why I suggested to use project specific environment variable to address your concerns about that the existing environment variable this PR using are global well known environment variable.

myakove: I prefer arg ant not os env.
when you create the client you can do 
`use_proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")`

myakove: use `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY`
Please update the readme as well.

sbahar619: Done.
I will add readme soon.

myakove:
@sbahar619 Please do no use force push next time

ocp_resources/resource.py
```diff
@@ -115,8 +124,22 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = client_configuration or kubernetes.client.Configuration()
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") == "1":
```
myakove, line 129: No need to do `== "1"` any value is valid to make this True

ocp_resources/resource.py
```diff
@@ -95,6 +100,10 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        client_configuration (kubernetes.client.Configuration): The kubernetes.client.Configuration to set configs to.
+        use_proxy (bool): If True, automatically retrieves and sets the proxy from the environment
+              variables (`HTTPS_PROXY` or `HTTP_PROXY`). If proxy is not found, raises an error.
+              Default is False, meaning no proxy is used.
```
myakove, line 106: `Default is False, meaning no proxy is used.` no need to specify this.

sbahar619:
> @sbahar619 Please do no use force push next time

Got it, from now on no force push.

README.md
```diff
@@ -93,6 +93,16 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
   export OPENSHIFT_PYTHON_WRAPPER_HASH_LOG_DATA="false"
   ```
 
+## Dynamic Proxy Enablement
+
+To dynamically enable proxy configuration:
+
+Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=1.
```
myakove, line 100:
```suggestion
Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>.
```

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
```diff
@@ -95,6 +100,10 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        client_configuration (kubernetes.client.Configuration): The kubernetes.client.Configuration to set configs to.
+        use_proxy (bool): If True, automatically retrieves and sets the proxy from the environment
```
myakove, line 104: retrieves HTTPS_PROXY or HTTP_PROXY from os environment and use it

myakove (changes requested):

ocp_resources/resource.py
sbahar619: ack

sbahar619: Done.

ocp_resources/resource.py
sbahar619: Done.

README.md
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
```diff
@@ -95,6 +100,8 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        client_configuration (kubernetes.client.Configuration): The kubernetes.client.Configuration to set configs to.
```
myakove, line 103: `client_configuration` should be full configuration from the user, it shouldn't be related to the `proxy`.
If the user pass `client_configuration` I expect it to be with proxy if this is what the user wanted.

Either remove it and only init `client_configuration` if user need proxy or keep them apart.

Did you test the flow with and without proxy?

ocp_resources/resource.py
```diff
@@ -115,8 +122,22 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = client_configuration or kubernetes.client.Configuration()
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
```
myakove, line 127:
```suggestion
        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
```

myakove (changes requested):

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Done.
Yes, tested with and without proxy.

ocp_resources/resource.py
```diff
@@ -115,8 +120,22 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kubernetes.client.Configuration()
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
+            use_proxy = True
```
myakove, line 126: `use_proxy = True` is not needed

ocp_resources/resource.py
```diff
@@ -115,8 +120,22 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kubernetes.client.Configuration()
```
myakove, line 123: Should be done only if proxy

myakove (changes requested):

myakove (changes requested):

ocp_resources/resource.py
```diff
@@ -95,6 +99,7 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        use_proxy (bool): If True, retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
```
rnetser, line 102: Please add that if  `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` is set, it will override `use_proxy`

rnetser (changes requested):

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
```diff
@@ -115,8 +120,22 @@ def get_client(
...
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kubernetes.client.Configuration()
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
+            use_proxy = True
+
+        if use_proxy:
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError("Proxy not found. Please set the HTTPS_PROXY or HTTP_PROXY environment variable.")
```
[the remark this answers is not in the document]
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Done.

sbahar619:
@myakove 
I forced push by mistake, I added the commits changes separately.

ocp_resources/resource.py
sbahar619: Done.

README.md
```diff
@@ -93,6 +93,16 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
   export OPENSHIFT_PYTHON_WRAPPER_HASH_LOG_DATA="false"
   ```
 
+## Dynamic Proxy Enablement
+
+To dynamically enable proxy configuration:
+
+Set the environment variable OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>.
```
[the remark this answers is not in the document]
sbahar619: Done.

README.md
```diff
@@ -93,6 +93,16 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
   export OPENSHIFT_PYTHON_WRAPPER_HASH_LOG_DATA="false"
   ```
 
+## Dynamic Proxy Enablement
```
myakove, line 96: Please remove the word `Dynamic`
Please specify that this is for the client.

myakove (changes requested):

README.md
sbahar619: Done.

ocp_resources/resource.py
```diff
@@ -115,8 +121,25 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = None
```
rnetser, line 124: `client_configuration` can be passed via `kwargs` , this will overwrite what the user sends.
set only if not passed

rnetser (changes requested):

ocp_resources/resource.py
sbahar619: Thanks, done.

ocp_resources/resource.py
```diff
@@ -115,8 +121,25 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kwargs.get("client_configuration", None)
+
+        if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
```
dbasunag, line 126: This is not matching with the line 103. With or condition, if use_proxy is true, value of `os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY")` won't matter.  But as per the header 
```
If the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` is set, it will override the value of `use_proxy`.
```

ocp_resources/resource.py
```diff
@@ -115,8 +121,25 @@ def get_client(
...
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
+                    "Proxy configuration requested but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set"
+                )
+            LOGGER.info(f"Setting proxy in client configuration: {proxy}")
```
dbasunag, line 132: This line printing `proxy` is not going to give extra data. You won't reach unless proxy has a `True` value.

dbasunag (changes requested):

ocp_resources/resource.py
sbahar619: The log is just a confirmation that the proxy is being set, but it doesn’t add value since we don’t log other configuration properties being set. It can be omitted to avoid unnecessary verbosity.
Done.

ocp_resources/resource.py
sbahar619: Right, thanks!
Done.

ocp_resources/resource.py
```diff
@@ -115,8 +121,24 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kwargs.get("client_configuration", None)
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
```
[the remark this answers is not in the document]
EdDev: The comment needs to be more accurate.

> If the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` is set, ...

 If the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` is set and not empty, ...

README.md
```diff
@@ -93,6 +93,21 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
   export OPENSHIFT_PYTHON_WRAPPER_HASH_LOG_DATA="false"
   ```
 
+## Proxy Enablement
+
+To enable proxy configuration for the client:
```
EdDev, line 98: It leave the reader clules about what this is for.
Can you provide a short explanation/overview for the purpose of this proxy?

ocp_resources/resource.py
```diff
@@ -95,6 +99,8 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        use_proxy (bool): If True, retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
+             If the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY` is set, it will override the value of `use_proxy`.
```
EdDev, line 103: Come to think of it, this is not a common pattern.
An explicit request from a caller is suppose to override any global env setting.

ocp_resources/resource.py
```diff
@@ -115,8 +121,24 @@ def get_client(
...
         return kubernetes.dynamic.DynamicClient(
-            client=kubernetes.config.new_client_from_config(config_file=config_file, context=context or None, **kwargs)
+            client=kubernetes.config.new_client_from_config(
+                config_file=config_file,
+                client_configuration=client_configuration,
+                context=context or None,
+                **kwargs,
+            )
         )
     except MaxRetryError:
```
EdDev, line 143: If the the new proxy is requested and we reached here, it means you need to fail as well. No?

ocp_resources/resource.py
```diff
@@ -115,8 +121,24 @@ def get_client(
...
+        client_configuration = kwargs.get("client_configuration", None)
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY") or use_proxy:
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
+                    "Proxy configuration requested but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set"
+                )
+            client_configuration = client_configuration or kubernetes.client.Configuration()
+            client_configuration.proxy = proxy
```
EdDev, line 133: If this configuration can come from the client config, why it is not under the test executor to set it correctly in the first place? Won't that make it transparent to the wrapper & tests?

sbahar619: The proxy configuration is handled at the wrapper level to avoid updating every client caller. This centralizes the proxy logic and ensures that all clients benefit from the proxy setting without needing individual updates. This approach provides flexibility while keeping the configuration transparent to the clients.

ocp_resources/resource.py
sbahar619: Thanks, so let's adjust the logic to use proxy first if use_proxy is set to True, or OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY env var set, which means use_proxy will always get precedence over the env.

ocp_resources/resource.py
sbahar619: I don't think so,
If we reached here, there is attempt to connect a client using incluster_config, which will also need the proxy configuration as the proxy was requested.

ocp_resources/resource.py
sbahar619: Done.

README.md
sbahar619: Done.

ocp_resources/resource.py
```diff
@@ -115,8 +122,24 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kwargs.get("client_configuration", None)
+
+        if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
```
myakove, line 127:
```suggestion
        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY", use_proxy):
```

ocp_resources/resource.py
```diff
@@ -115,8 +122,24 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        client_configuration = kwargs.get("client_configuration", None)
```
myakove, line 125: `None` is the default return, no need to specify it.
Better to do `client_configuration = kwargs.get("client_configuration", kubernetes.client.Configuration())`

ocp_resources/resource.py
myakove: I accepted this solution to make cnv-tests life easier for now, I expect them to refactor their code and do it right.
The right solution is to send `client_configuration` with the needed values.

ocp_resources/resource.py
```diff
@@ -115,8 +122,24 @@ def get_client(
...
+
+        if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
+                    "Proxy configuration requested but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set"
+                )
+            client_configuration = client_configuration or kubernetes.client.Configuration()
+            client_configuration.proxy = proxy
```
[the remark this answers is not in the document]
myakove: @sbahar619 please check coderabbitai comments as well.
FYI we do not want to validate proxy URL schema.

myakove (changes requested):

ocp_resources/resource.py
sbahar619: Ohhh... I misunderstood you.
You are right, if the user already defined a client_configuration, we shouldn't override it, the user needs to pre define the proxy with it.
I will update it, thanks!

sbahar619: Done.

ocp_resources/resource.py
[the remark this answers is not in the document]
sbahar619: @myakove 
I did, I can't see any additional value in the suggested changes:
Why to set default value to "" explicitly when the implicit default value is None and both works for the if statement condition?
We don't want to validate URL schema.
I adjusted the error log as recommend.

ocp_resources/resource.py
EdDev: Caller explicit request takes precedence over env setting, so this suggestion does not look right.

ocp_resources/resource.py
EdDev: I do not understand.
If you reached here, does it mean `client_configuration` is updated with the proxy or not?

ocp_resources/resource.py
```diff
@@ -115,15 +123,31 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+
+        if use_proxy or os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
```
dbasunag, line 128: No need to save

dbasunag (changes requested):

ocp_resources/resource.py
EdDev: Sorry, I still do not understand something here.
Can the client configuration be specified on the environment itself instead on the wrapper or tests?
(e.g. changing the kubeconfig file).

ocp_resources/resource.py
sbahar619: +1
@EdDev

ocp_resources/resource.py
sbahar619: Why?
Currently it is being called twice.

ocp_resources/resource.py
myakove: @sbahar619 You +1 @EdDev comment but it's against what you asked me for, to be able to run without any changes to your code base. 

If user_proxy is False by default and this is what takes precedence then the os env is useless.

You can do `use_proxy: bool|None = None` and ony then you can check if the user sent it to the function.

sbahar619: @myakove 
It's not accurate, the environment variable is not useless, I added this new use_proxy argument, and its false by default, but if the env is set, it will be respected, it is as an addition to the argument.
It will help us and avoid updating the caller with the new argument.
It doesn't mean the env take precedence over the argument, as if the argument is True, the env won't matter at all, it only affects when argument is false.

EdDev: @sbahar619 , I think that what @myakove is trying to say is something like this:
- If the caller has an opinion, it will explicitly say so, either to enable or disable the proxy.
- If the caller does not care, it will not specify it at all.
- If the called does not care, look at the env var and take a decision based on its input, if it exists.

sbahar619: @EdDev @myakove 
let's clarify the possible combinations of the use_proxy argument and the environment variable, which result in four scenarios:
1. use_proxy is False (default) or explicitly set by the user, and the environment variable is not set: Proxy is disabled.
2. use_proxy is False (default) or explicitly set by the user, and the environment variable is set: Proxy is enabled.
3. use_proxy is True, and the environment variable is not set: Proxy is enabled.
4. use_proxy is True, and the environment variable is set: Proxy is enabled.

In scenario 2, the environment variable overrides the use_proxy argument when it's False, but since that's the default behavior, I understand it's not ideal. However, this approach is a tradeoff— it allows the environment variable to take precedence when the user hasn't explicitly set use_proxy to True, while still respecting the user's intent when use_proxy is set to True.

Do you think we should adjust this logic to give the environment variable full precedence?

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Yes, the client configuration can be specified via the environment by setting the KUBECONFIG environment variable, which points to the kubeconfig file. This allows you to avoid modifying the wrapper or tests directly, as the configuration will be automatically loaded from the specified kubeconfig file when you call load_kube_config().

However, not all properties in the kubeconfig file are supported by load_kube_config(). For instance, the proxy-url property is not populated by default. In our PR, I set the proxy directly in the client configuration. Additionally, there's an upstream PR ([#2182](https://github.com/kubernetes-client/python/pull/2182)) which I also added to this PR description already, to add support for populating the proxy-url in the client configuration from kubeconfig. This should help address the issue and allow the proxy-url to be automatically loaded in future releases.
We can contribute this upstream PR.

sbahar619:
> @sbahar619 , can you please respond to [#2291 (comment)](https://github.com/RedHatQE/openshift-python-wrapper/pull/2291#discussion_r1932105802) ? Github is hiding stuff and makes it extremely hard to find them :/

Done:
https://github.com/RedHatQE/openshift-python-wrapper/pull/2291#discussion_r1933402128

ocp_resources/resource.py
EdDev: @sbahar619 , the scenarios are per caller intent, not per raw combinations.
The scenarios I believe @myakove looks for, and I as well, is the one I shared prev.
Using the boolean as it appears now, cannot express the caller intention of "I do not care or know".

sbahar619: @EdDev 
Let's be practical. Are you suggesting that "I do not care or know" should be represented as a None value, which would be the default? This way, the user can explicitly set use_proxy to True or False, and it will take precedence, determining whether the proxy is enabled or not. If use_proxy is None, the environment variable will be checked to decide whether to enable the proxy.

EdDev: This is what @myakove hinted you to do. At least that is my interpretation of [this](https://github.com/RedHatQE/openshift-python-wrapper/pull/2291#discussion_r1932335166) comment.

sbahar619: Ohhhh, sorry for the delayed understanding, I will update it soon, thanks!

sbahar619: Done.

README.md
EdDev: Please resolve this thread

EdDev (approved):

ocp_resources/resource.py
```diff
@@ -92,6 +96,11 @@ def get_client(
     Args:
         config_file (str): path to a kubeconfig file.
         config_dict (dict): dict with kubeconfig configuration.
+        use_proxy (bool, default=None):
+            - `True`: retrieves HTTPS_PROXY or HTTP_PROXY from OS environment and use it.
+            - `False`: Disables proxy.
```
dbasunag, line 101: If I pass None and there is no OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY, does not that mean proxy would not be used? Do you need to now set `client_configuration.proxy` to False too? It seems a bit overcomplicated.

sbahar619: No, if you pass use_proxy=True and env is not set, it doesn't matter, you already enable the proxy.
The env only apply when the use_proxy=None and when its set it enables proxy.
If the user set use_proxy=false, the env won't matter.

EdDev: The documentation looks ok to me, however, if there is a way to better phrase it, please suggest something.

What is overcomplicated? I find it trivial that if a caller expresses intent it should take precedence over some “default”, in this case, to read it from the env.

ocp_resources/resource.py
```diff
@@ -112,15 +122,30 @@ def get_client(
...
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+
+        if use_proxy or (use_proxy is None and os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY")):
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
+                    "Proxy configuration is enabled but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set."
+                )
+            if not kwargs.get("client_configuration"):
```
myakove, line 132: Why do you need to check it again? you already set it in line 116

myakove (changes requested):

ocp_resources/resource.py
```diff
@@ -76,7 +76,11 @@ def _get_api_version(dyn_client: DynamicClient, api_group: str, kind: str) -> st
 
 
 def get_client(
-    config_file: str = "", config_dict: dict[str, Any] | None = None, context: str = "", **kwargs: Any
+    config_file: str = "",
+    config_dict: dict[str, Any] | None = None,
+    use_proxy: bool | None = None,
```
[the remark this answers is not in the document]
dbasunag: +1

ocp_resources/resource.py
dbasunag: It is not about phrasing the doc. adding use_proxy is overcomplicating it. I prefer @myakove's suggestion of using os env variables instead of use_proxy param

ocp_resources/resource.py
sbahar619: If we don’t check, we risk overriding a user-provided client_configuration, including their proxy settings. If a user explicitly disables the proxy in their configuration, I don't think we should forcefully enable it. This check ensures that we respect user-defined settings.
Do you think we do need to enable it even if client_configuration was passed by a user?

ocp_resources/resource.py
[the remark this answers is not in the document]
sbahar619: Done.

ocp_resources/resource.py
myakove: If the user set both it's a bad call, we do not need to check twice.

myakove (changes requested):

ocp_resources/resource.py
myakove: You should do:
```
if client_configuration.proxy and client_configuration.proxy != proxy:
    raise
```

ocp_resources/resource.py
```diff
@@ -112,15 +116,29 @@ def get_client(
...
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
+                    "Proxy configuration is enabled but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set."
+                )
+            client_configuration.proxy = proxy
```
myakove, line 126: You should check here if `client_configuration` contain proxy and compare it to the one we got from the os env.
```
if client_configuration.proxy and client_configuration.proxy != proxy:
    raise .......
```

myakove (changes requested):

ocp_resources/resource.py
myakove: @EdDev can this be resolved?

ocp_resources/resource.py
sbahar619: Done.
Example:
```
INTERNALERROR>   File "/home/sbahar/projects/cnv/openshift-python-wrapper/ocp_resources/resource.py", line 127, in get_client
INTERNALERROR>     raise ValueError(
INTERNALERROR> ValueError: Conflicting proxy settings: client_configuration.proxy=http://api.cool.com, but the environment variable 'OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY' defines proxy as http://cnvqe-30.lab.eng.tlv2.redhat.com:3128.
```

myakove: Thanks, please add test for this.
https://github.com/RedHatQE/openshift-python-wrapper/blob/main/tests/test_resources.py

sbahar619: Done and verified.

tests/test_resources.py
```diff
@@ -103,3 +104,17 @@ def test_resource_context_manager_exit(self, client):
...
                 pass
+
+    def test_proxy_conflict_raises_value_error(self, monkeypatch):
+        monkeypatch.setenv(name="OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY", value="1")
+        monkeypatch.setenv(name="HTTPS_PROXY", value="http://env-proxy.com")
+
+        client_configuration = kubernetes.client.Configuration()
+        client_configuration.proxy = "http://not-env-proxy.com"
+
+        with pytest.raises(ValueError) as exc_info:
```
rnetser, line 115: please use `pytest.raises`  `match` option to verify the error message

ocp_resources/resource.py
```diff
@@ -112,15 +116,34 @@ def get_client(
         # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
         # is populated during import which comes before setting the variable in code.
         config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
```
rnetser, line 123: please add test

ocp_resources/resource.py
```diff
@@ -112,15 +116,34 @@ def get_client(
...
+                    "Proxy configuration is enabled but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set."
+                )
+            if client_configuration.proxy and client_configuration.proxy != proxy:
+                raise ValueError(
+                    f"Conflicting proxy settings: client_configuration.proxy={client_configuration.proxy}, "
+                    f"but the environment variable 'OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY' defines proxy as {proxy}."
+                )
+            client_configuration.proxy = proxy
```
rnetser, line 131: move `client_configuration.proxy = proxy` before the `if` and re-use

rnetser (changes requested):

tests/test_resources.py
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Done.

ocp_resources/resource.py
sbahar619: Moving client_configuration.proxy = proxy before the if would overwrite the user-set proxy, preventing conflict detection. The current order ensures we check for conflicts before assigning the proxy. If the suggestion meant something else, I might have misunderstood—can you clarify?

ocp_resources/resource.py
EdDev: Yes

ocp_resources/resource.py
myakove: No need to reuse this.

myakove (approved):

README.md
```diff
@@ -93,6 +93,23 @@ export OPENSHIFT_PYTHON_WRAPPER_LOG_LEVEL=<LOG_LEVEL> # can be: "DEBUG", "INFO",
...
   ```
 
+## Proxy Enablement
+
+This configuration allows the client to route traffic through a specified proxy server.
+It can be enabled via the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY`.
+
+To enable proxy configuration for the client:
+
+1. Set the environment variable `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=<any value>`
```
dbasunag, line 103: Can you give example? What are the valid options? or is the goal to ensure that it is set?
e.g. Currently if I set it to `OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY=False`, the code would still set proxy provided HTTPS_PROXY or HTTP_PROXY exists.

ocp_resources/resource.py
```diff
@@ -112,15 +116,34 @@ def get_client(
...
+        if os.environ.get("OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY"):
+            proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+            if not proxy:
+                raise ValueError(
+                    "Proxy configuration is enabled but neither HTTPS_PROXY nor HTTP_PROXY environment variables are set."
+                )
+            if client_configuration.proxy and client_configuration.proxy != proxy:
+                raise ValueError(
+                    f"Conflicting proxy settings: client_configuration.proxy={client_configuration.proxy}, "
+                    f"but the environment variable 'OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY' defines proxy as {proxy}."
```
dbasunag, line 129: This message is misleading the value of proxy is not coming from OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY

README.md
myakove: Any value is ok and makes sense

dbasunag: @myakove Should that be mentioned that you can't disable proxy using OPENSHIFT_PYTHON_WRAPPER_CLIENT_USE_PROXY? One needs to unset this environment variable? Currently it is not clear.

ocp_resources/resource.py
myakove: I'm going to merge the PR and I will fix the error in follow up PR

PR #2334, opened 2025-02-26
Rename NETWORK_READY condition to NETWORK_CREATED

sbahar619:
##### Short description:
Rename NETWORK_READY condition to NETWORK_CREATED
##### More details:
The condition NETWORK_READY is now NETWORK_CREATED to align with upstream changes in:
ovn-kubernetes PR[ #4884](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4884).

The method wait_for_network_ready is also renamed
to wait_for_network_created for clarity.

This ensures consistency with ovn-kubernetes and
avoids ambiguity in network status handling.
## Summary by CodeRabbit

- **Refactor**
  - Refined the network initialization process to ensure that status messages and error feedback more accurately reflect the network creation phase.

### DIFF
--- a/ocp_resources/user_defined_network.py
+++ b/ocp_resources/user_defined_network.py
@@ -68,18 +68,18 @@ class UserDefinedNetwork(NamespacedResource):
 
     # End of generated code
 
-    def wait_for_network_ready(self, timeout: int = 30) -> None:
+    def wait_for_network_created(self, timeout: int = 30) -> None:
         """
-        Wait for the network to be ready.
+        Wait for the network to be created.
 
         Args:
             timeout (int, optional): Maximum time to wait in seconds. Defaults to 30.
 
         Raises:
-            TimeoutExpiredError: If the network is not ready within the specified timeout.
+            TimeoutExpiredError: If the network is not created within the specified timeout.
         """
         self.wait_for_condition(
-            condition=self.Condition.NETWORK_READY,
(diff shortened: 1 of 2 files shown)


ocp_resources/user_defined_network.py
```diff
@@ -68,18 +68,18 @@ def to_dict(self) -> None:
 
     # End of generated code
 
-    def wait_for_network_ready(self, timeout: int = 30) -> None:
+    def wait_for_network_created(self, timeout: int = 30) -> None:
```
myakove, line 71: again, this will break anyone who use the code.

myakove (changes requested):

ocp_resources/user_defined_network.py
EdDev: Yes, you break existing usage. Do not change the name.

This change also is in collision of https://github.com/RedHatQE/openshift-virtualization-tests/pull/294/commits/6594c27c1682193222cd3caf7158a81874357674
There you used the `NetworkAllocationSucceeded` condition.

EdDev (changes requested):

ocp_resources/user_defined_network.py
sbahar619: So, we can keep it, I don't see a reason to add deprecation warning, because there isn't any user argument to warn the user about.
Without changing the method name, the context will be wrong.

PR #2347, opened 2025-03-17
fix: Avoid race when removing interfaces via NNCP

yossisegev:
Remoivng an interface that was created using an NNCP, is done by editing the same NNCP. This sometimes resulted in a race, in which the NNCP success status actually presented the prvious status, leading to deleting the NNCP before the configuration was completed, leaving hanging interfaces in the cluster nodes, with node native interfaces occupied as the ports of these tests-created interfaces. A recent PR made this failed flow to always occur. This PR aims to assure that the timestamp of the AVAIALBLE status is updated for the recent change (the interface removal) and not the previous change (setup or modification).
This PR is based on the fix that was presented in
https://github.com/RedHatQE/openshift-virtualization-tests/pull/430.

## Summary by CodeRabbit

- **New Features**
	- Improved network configuration updates now include an added verification step that ensures changes are fully applied, enhancing reliability during updates.
- **Chores**
	- Updated Flake8 configuration to exclude the `datetime` function from certain checks.

### DIFF
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -333,7 +333,7 @@ class NodeNetworkConfigurationPolicy(Resource):
         if initial_success_status_time:
             self._wait_for_nncp_status_update(initial_transition_time=initial_success_status_time)
 
-    def _get_last_successful_transition_time(self) -> str:
+    def _get_last_successful_transition_time(self) -> str | None:
         for condition in self.instance.status.conditions:
             if (
                 condition["type"] == self.Conditions.Type.AVAILABLE
@@ -341,6 +341,7 @@ class NodeNetworkConfigurationPolicy(Resource):
                 and condition["reason"] == self.Conditions.Reason.SUCCESSFULLY_CONFIGURED
             ):
                 return condition["lastTransitionTime"]
+        return None
 
     @retry(
         wait_timeout=TIMEOUT_1MINUTE,
@@ -355,6 +356,7 @@ class NodeNetworkConfigurationPolicy(Resource):
                 and datetime.strptime(condition["lastTransitionTime"], date_format) > formatted_initial_transition_time
             ):
                 return True
+        return False
 
     @property
     def status(self):
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -1,8 +1,9 @@
 import re
+from datetime import datetime
 
 from kubernetes.dynamic.exceptions import ConflictError
 
-from ocp_resources.utils.constants import TIMEOUT_4MINUTES
+from ocp_resources.utils.constants import TIMEOUT_1MINUTE, TIMEOUT_4MINUTES, TIMEOUT_5SEC
 from ocp_resources.exceptions import NNCPConfigurationFailed
 from ocp_resources.node import Node
 from ocp_resources.node_network_configuration_enactment import (
@@ -10,7 +11,7 @@ from ocp_resources.node_network_configuration_enactment import (
 )
 from ocp_resources.node_network_state import NodeNetworkState
 from ocp_resources.resource import Resource, ResourceEditor
-from timeout_sampler import TimeoutExpiredError, TimeoutSampler, TimeoutWatch
+from timeout_sampler import TimeoutExpiredError, TimeoutSampler, TimeoutWatch, retry
 
 IPV4_STR = "ipv4"
 IPV6_STR = "ipv6"
@@ -447,3 +448,33 @@ class NodeNetworkConfigurationPolicy(Resource):
             for nnce_cond in nnce.instance.status.conditions:
                 if nnce_cond.type == "Failing" and nnce_cond.status == Resource.Condition.Status.TRUE:
                     yield nnce
+
+    def _get_nncp_configured_last_transition_time(self):
+        for condition in self.instance.status.conditions:
+            if (
+                condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+                and condition["status"] == "True"
+                and condition["reason"] == NodeNetworkConfigurationPolicy.Conditions.Reason.SUCCESSFULLY_CONFIGURED
+            ):
+                return condition["lastTransitionTime"]
+
+    @retry(
+        wait_timeout=TIMEOUT_1MINUTE,
+        sleep=TIMEOUT_5SEC,
+    )
+    def _wait_for_nncp_with_different_transition_time(self, initial_transition_time):
+        date_format = "%Y-%m-%dT%H:%M:%SZ"
+        for condition in self.instance.get("status", {}).get("conditions", []):
+            if (
+                condition
+                and condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+                and datetime.strptime(condition["lastTransitionTime"], date_format)
+                > datetime.strptime(initial_transition_time, date_format)
+            ):
+                return True
+        return False
+
+    def update(self, resource_dict=None):
+        initial_transition_time = self._get_nncp_configured_last_transition_time()
+        super().update(resource_dict=resource_dict)
+        self._wait_for_nncp_with_different_transition_time(initial_transition_time=initial_transition_time)
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -323,21 +323,22 @@ class NodeNetworkConfigurationPolicy(Resource):
         if self.ports:
             self.add_ports()
 
-        initial_transition_time = self._get_nncp_configured_last_transition_time()
+        initial_success_status_time = self._get_last_successful_transition_time()
         ResourceEditor(
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
-        # If the setup NNCP failed - we don't need to verify its teardown status time is updated.
-        if initial_transition_time:
-            self._wait_for_nncp_status_update(initial_transition_time=initial_transition_time)
+        # If the NNCP failed on setup, then at this point it's not in AVAILABLE status, so the next time it will be in
+        # AVAILABLE status will necessarily be the updated one.
+        if initial_success_status_time:
+            self._wait_for_nncp_status_update(initial_transition_time=initial_success_status_time)
 
-    def _get_nncp_configured_last_transition_time(self):
+    def _get_last_successful_transition_time(self) -> str:
         for condition in self.instance.status.conditions:
             if (
-                condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
-                and condition["status"] == "True"
-                and condition["reason"] == NodeNetworkConfigurationPolicy.Conditions.Reason.SUCCESSFULLY_CONFIGURED
+                condition["type"] == self.Conditions.Type.AVAILABLE
+                and condition["status"] == Resource.Condition.Status.TRUE
+                and condition["reason"] == self.Conditions.Reason.SUCCESSFULLY_CONFIGURED
             ):
                 return condition["lastTransitionTime"]
 
@@ -345,14 +346,15 @@ class NodeNetworkConfigurationPolicy(Resource):
         wait_timeout=TIMEOUT_1MINUTE,
         sleep=TIMEOUT_5SEC,
     )
-    def _wait_for_nncp_status_update(self, initial_transition_time):
+    def _wait_for_nncp_status_update(self, initial_transition_time: str) -> bool:
         date_format = "%Y-%m-%dT%H:%M:%SZ"
         formatted_initial_transition_time = datetime.strptime(initial_transition_time, date_format)
-        return any(
-            condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
-            and datetime.strptime(condition["lastTransitionTime"], date_format) > formatted_initial_transition_time
-            for condition in self.instance.get("status", {}).get("conditions", [])
-        )
+        for condition in self.instance.get("status", {}).get("conditions", []):
+            if (
+                condition["type"] == self.Conditions.Type.AVAILABLE
+                and datetime.strptime(condition["lastTransitionTime"], date_format) > formatted_initial_transition_time
+            ):
+                return True
 
     @property
     def status(self):
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -464,12 +464,13 @@ class NodeNetworkConfigurationPolicy(Resource):
     )
     def _wait_for_nncp_with_different_transition_time(self, initial_transition_time):
         date_format = "%Y-%m-%dT%H:%M:%SZ"
+        formatted_initial_transition_time = datetime.strptime(initial_transition_time, date_format)
         for condition in self.instance.get("status", {}).get("conditions", []):
             if (
                 condition
                 and condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
                 and datetime.strptime(condition["lastTransitionTime"], date_format)
-                > datetime.strptime(initial_transition_time, date_format)
+                > formatted_initial_transition_time
             ):
                 return True
         return False
--- a/ocp_resources/node_network_configuration_policy.py
+++ b/ocp_resources/node_network_configuration_policy.py
@@ -323,15 +323,14 @@ class NodeNetworkConfigurationPolicy(Resource):
         if self.ports:
             self.add_ports()
 
-        # The time-stamp that is returned by _get_last_successful_transition_time() will change after the call to
-        # update(), therefore it must be fetched and stored before, and compared with the new time-stamp after.
+        # The current time-stamp of the NNCP's available status will change after the NNCP is updated, therefore
+        # it must be fetched and stored before the update, and compared with the new time-stamp after.
         initial_success_status_time = self._get_last_successful_transition_time()
         ResourceEditor(
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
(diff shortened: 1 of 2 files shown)


ocp_resources/node_network_configuration_policy.py
```diff
@@ -447,3 +448,33 @@ def _get_failed_nnce(self):
...
+        sleep=TIMEOUT_5SEC,
+    )
+    def _wait_for_nncp_with_different_transition_time(self, initial_transition_time):
+        date_format = "%Y-%m-%dT%H:%M:%SZ"
+        for condition in self.instance.get("status", {}).get("conditions", []):
+            if (
+                condition
+                and condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+                and datetime.strptime(condition["lastTransitionTime"], date_format)
+                > datetime.strptime(initial_transition_time, date_format)
```
sbahar619, line 472: You can avoids repeatedly parsing initial_transition_time inside the loop, by moving it out the loop and save it:
initial_time = datetime.strptime(initial_transition_time, date_format)

ocp_resources/node_network_configuration_policy.py
```diff
@@ -447,3 +448,33 @@ def _get_failed_nnce(self):
...
+            ):
+                return condition["lastTransitionTime"]
+
+    @retry(
+        wait_timeout=TIMEOUT_1MINUTE,
+        sleep=TIMEOUT_5SEC,
+    )
+    def _wait_for_nncp_with_different_transition_time(self, initial_transition_time):
+        date_format = "%Y-%m-%dT%H:%M:%SZ"
+        for condition in self.instance.get("status", {}).get("conditions", []):
```
sbahar619, line 467: You can use any statement and eliminates the need for an explicit for loop with return True.
return any(...)

yossisegev: I think that wrapping the condition criteria in an `any` statement will make things less clear, and it will be harder to understand what is being checked here.
And performance-wise it won't give us anything, because `any` still iterates over all the list members - just like a `for` loop.
WDYT?

sbahar619: With any you also achieve less code statements and its more Pythonic, but it's a nit comment, you will choose how to implement it.
I am OK with both approaches.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -447,3 +448,34 @@ def _get_failed_nnce(self):
...
+                condition
+                and condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+                and datetime.strptime(condition["lastTransitionTime"], date_format)
+                > formatted_initial_transition_time
+            ):
+                return True
+        return False
+
+    def update(self, resource_dict=None):
+        initial_transition_time = self._get_nncp_configured_last_transition_time()
```
sbahar619, line 479: Please move it close as possible to the caller.

redhat-qe-bot1:
not adding reviewer EdDev by user comment, EdDev is not part of contributers

ocp_resources/node_network_configuration_policy.py
sbahar619, line 479: You can call it directly in:
self._wait_for_nncp_with_different_transition_time(initial_transition_time=self._get_nncp_configured_last_transition_time())

The parameter name initial_transition_time already giving the context of the argument.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -447,3 +448,34 @@ def _get_failed_nnce(self):
...
+            if (
+                condition
+                and condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+                and datetime.strptime(condition["lastTransitionTime"], date_format)
+                > formatted_initial_transition_time
+            ):
+                return True
+        return False
+
+    def update(self, resource_dict=None):
```
sbahar619, line 478: Why not calling this logic here:
https://github.com/RedHatQE/openshift-python-wrapper/blob/295331783628b5f325f71a9ee4dea5f2b1004d23/ocp_resources/node_network_configuration_policy.py#L308
So whenever a cleanup will be executed, you will wait for new timestamp.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -447,3 +448,34 @@ def _get_failed_nnce(self):
...
+                and condition["status"] == "True"
+                and condition["reason"] == NodeNetworkConfigurationPolicy.Conditions.Reason.SUCCESSFULLY_CONFIGURED
+            ):
+                return condition["lastTransitionTime"]
+
+    @retry(
+        wait_timeout=TIMEOUT_1MINUTE,
+        sleep=TIMEOUT_5SEC,
+    )
+    def _wait_for_nncp_with_different_transition_time(self, initial_transition_time):
```
sbahar619, line 465: I think better name will be something like:
_wait_for_nncp_status_update

ocp_resources/node_network_configuration_policy.py
yossisegev: I changed it to `any()`.
I must say I don't see how it makes the code clearer/nittier/whateverier, but if that's what you prefer - so be it.

ocp_resources/node_network_configuration_policy.py
yossisegev: If I do that. it will miss the purpose of this PR :) 
Storing `initial_transition_time` must be done before calling `update()`.

ocp_resources/node_network_configuration_policy.py
yossisegev: Why does it matter if it's in `clean_up()` or in `_absent_interface()`?

ocp_resources/node_network_configuration_policy.py
sbahar619: You are absolutely right 🤝😎

ocp_resources/node_network_configuration_policy.py
```diff
@@ -326,6 +327,35 @@ def _absent_interface(self):
...
+    def _wait_for_nncp_status_update(self, initial_transition_time):
+        date_format = "%Y-%m-%dT%H:%M:%SZ"
+        formatted_initial_transition_time = datetime.strptime(initial_transition_time, date_format)
+        if any(
+            condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+            and datetime.strptime(condition["lastTransitionTime"], date_format) > formatted_initial_transition_time
+            for condition in self.instance.get("status", {}).get("conditions", [])
+        ):
+            return True
+        return False
```
sbahar619, line 357: After adding the any statement, the return statements are redundant, you can simply return:
```
return any(
            condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
            and datetime.strptime(condition["lastTransitionTime"], date_format) > formatted_initial_transition_time
            for condition in self.instance.get("status", {}).get("conditions", [])
        )
```

ocp_resources/node_network_configuration_policy.py
sbahar619: You’re not using this directly in _absent_interface; rather, _absent_interface calls the update method, which you implemented.

It would be more accurate to move this logic to the cleanup method since the issue occurs during teardown—unless there are cases where the NNCP is updated to remove interfaces without deleting the NNCP object itself.

Additionally, implementing this in update forces the NNCP to reach a new status, but what if we only update metadata labels? In that case, the update would fail to wait for a new status… could this be a bug?

ocp_resources/node_network_configuration_policy.py
```diff
@@ -326,6 +327,35 @@ def _absent_interface(self):
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
+    def update(self, resource_dict=None):
+        initial_transition_time = self._get_nncp_configured_last_transition_time()
```
sbahar619, line 331: This can be None, if the condition is not met yet for example, and then all the workflow will raise exception and fail somewhere.
We need to handle it.

yossisegev:
@myakove 
coderabbit complains about:
```ocp_resources/node_network_configuration_policy.py:350:45: FCN001: [datetime] function should be called with keywords arguments. value: initial_transition_time (line:350 column:62)value: date_format (line:350 column:87)```
But when I call `strptime` with these explicit argument names (`date_string` and `format`, based on the [function documentataion](https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime)) it just fails.
Can we just ignore the rabbit here?
Thank you

File: ocp_resources/node_network_configuration_policy.py [hunk 8]
```diff
@@ -322,10 +323,37 @@ def _absent_interface(self):
...
         ResourceEditor(
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
+        # If the setup NNCP failed - we don't need to verify its teardown status time is updated.
+        if initial_transition_time:
+            self._wait_for_nncp_status_update(initial_transition_time=initial_transition_time)
+
+    def _get_nncp_configured_last_transition_time(self):
```
myakove, line 335: please add typing.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,37 @@ def _absent_interface(self):
...
+                and condition["status"] == "True"
+                and condition["reason"] == NodeNetworkConfigurationPolicy.Conditions.Reason.SUCCESSFULLY_CONFIGURED
+            ):
+                return condition["lastTransitionTime"]
+
+    @retry(
+        wait_timeout=TIMEOUT_1MINUTE,
+        sleep=TIMEOUT_5SEC,
+    )
+    def _wait_for_nncp_status_update(self, initial_transition_time):
```
myakove, line 348: please add typing.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,37 @@ def _absent_interface(self):
...
+        date_format = "%Y-%m-%dT%H:%M:%SZ"
+        formatted_initial_transition_time = datetime.strptime(initial_transition_time, date_format)
+        return any(
```
myakove, line 351: You use here `any` but it will be always only one match right?
if so please operate only on the type you need (similar to your above code for example)

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
sbahar619, line 335: One possible scenario is that this method returns None because the condition reason is ConfigurationProgressing. This state is typically temporary and recoverable, so it might be worth waiting for the condition to change before proceeding.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,37 @@ def _absent_interface(self):
...
+        # If the setup NNCP failed - we don't need to verify its teardown status time is updated.
+        if initial_transition_time:
+            self._wait_for_nncp_status_update(initial_transition_time=initial_transition_time)
+
+    def _get_nncp_configured_last_transition_time(self):
+        for condition in self.instance.status.conditions:
+            if (
+                condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
+                and condition["status"] == "True"
+                and condition["reason"] == NodeNetworkConfigurationPolicy.Conditions.Reason.SUCCESSFULLY_CONFIGURED
```
sbahar619, line 340:
```suggestion
                condition["type"] == self.Conditions.Type.AVAILABLE
                and condition["status"] == Resource.Condition.Status.TRUE
                and condition["reason"] == self.Conditions.Reason.SUCCESSFULLY_CONFIGURED
```

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,37 @@ def _absent_interface(self):
...
+
+    @retry(
+        wait_timeout=TIMEOUT_1MINUTE,
+        sleep=TIMEOUT_5SEC,
+    )
+    def _wait_for_nncp_status_update(self, initial_transition_time):
+        date_format = "%Y-%m-%dT%H:%M:%SZ"
+        formatted_initial_transition_time = datetime.strptime(initial_transition_time, date_format)
+        return any(
+            condition["type"] == NodeNetworkConfigurationPolicy.Conditions.Type.AVAILABLE
```
sbahar619, line 352:
```suggestion
            condition["type"] == self.Conditions.Type.AVAILABLE
```

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,37 @@ def _absent_interface(self):
         if self.ports:
             self.add_ports()
 
+        initial_transition_time = self._get_nncp_configured_last_transition_time()
         ResourceEditor(
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
+        # If the setup NNCP failed - we don't need to verify its teardown status time is updated.
```
sbahar619, line 331: Are you referring to NNCP failing as FailedToConfigure? If so, it could also be in states like ConfigurationProgressing or NoMatchingNode, which makes this comment unclear.

ocp_resources/node_network_configuration_policy.py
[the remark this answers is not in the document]
yossisegev: I assumed that "configured" reflects a successful NNCP, but if you think this is clearer then no problem, I changed it (and removed "nncp").

yossisegev: This function is called before editing a "stable" NNCP, meaning - the NNCP was previously configured successfully, and at this moment it is necessarily not in progressing state, but rather necessarily in SuccessfullyConfigured state.

ocp_resources/node_network_configuration_policy.py
Anatw, line 351: You know that is only a single Available condition, why not access it directly? Something like:
    conditions = self.instance.get("status", {}).get("conditions", [])
    if conditions and conditions[0]["type"] == "Available":
        last_transition_time = datetime.strptime(conditions[0]["lastTransitionTime"], date_format)
        return last_transition_time > formatted_initial_transition_time

ocp_resources/node_network_configuration_policy.py
sbahar619: Right, thanks.

ocp_resources/node_network_configuration_policy.py
yossisegev: @sbahar619 Please note, as you requested this change:
`any` doesn't fit here, because there is only one state that is valid (if `type==AVAILABLE` and `lastTransitionTime > formatted_initial_time`). There is not a set of conditions, were it's enough that one of them is true.

ocp_resources/node_network_configuration_policy.py
yossisegev: I changed the comment and the name of the var, I hope it's clearer now.

ocp_resources/node_network_configuration_policy.py
yossisegev: You cannot assume that a target member is located in a specific index of a list.
You cannot access `conditions[0]' directly and assume the this is the condition that you are interested in. You must iterate over all the conditions in the list and find the one that you are interested in.

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,39 @@ def _absent_interface(self):
         if self.ports:
             self.add_ports()
 
+        initial_success_status_time = self._get_last_successful_transition_time()
         ResourceEditor(
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
+        # If the NNCP failed on setup, then at this point it's not in AVAILABLE status, so the next time it will be in
```
sbahar619, line 331: If NNCP fail on setup, it will not reach this code at all:
https://github.com/RedHatQE/openshift-python-wrapper/blob/e9a1776878bddff0dec44aa1929817b4ca7701ba/ocp_resources/node_network_configuration_policy.py#L295

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,41 @@ def _absent_interface(self):
...
+        # AVAILABLE status will necessarily be the updated one.
+        if initial_success_status_time:
```
myakove, line 333:
```suggestion
        if initial_success_status_time := self._get_last_successful_transition_time()
```
and remove line 326

myakove (changes requested):

ocp_resources/node_network_configuration_policy.py
yossisegev: I can't.
If I do that, it will miss the purpose of this PR :)
Storing initial_transition_time must be done before calling `update()`.

myakove: Please add a comment above line 326 about this

ocp_resources/node_network_configuration_policy.py
```diff
@@ -322,10 +323,43 @@ def _absent_interface(self):
...
 
+        # The time-stamp that is returned by _get_last_successful_transition_time() will chnage after the call to
+        # update(), therefore it must be fetched and stored before, and comapred with the new time-stamp after.
+        initial_success_status_time = self._get_last_successful_transition_time()
         ResourceEditor(
             patches={self: {"spec": {"desiredState": {"interfaces": self.desired_state["interfaces"]}}}}
         ).update()
 
+        # If the NNCP failed on setup, then at this point it's not in AVAILABLE status, so the next time it will be in
+        # AVAILABLE status will necessarily be the updated one.
```
sbahar619, line 334: nit
The current comments contain specific low-level details, such as function names, which may not be necessary. Since the function names already describe their purpose, we can simplify the comments to focus more on why this logic is needed rather than what it does.

For example:
```
# Store the current success timestamp before update
# to detect if the status changes afterward.  
...
# If NNCP setup fails, the next AVAILABLE status must be a new one.
```

sbahar619 (approved):
lgtm.

myakove (approved):

PR #2358, opened 2025-03-21
fix: Do not set signal when not in main thread

sarahbx:
`Resource.__enter__` calls signal which
cannot be called when in a different
thread from main.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved event handling stability by ensuring that specific operations are executed exclusively in the main thread. This update prevents potential misbehavior during concurrent operations, enhancing overall system reliability and providing a smoother, more predictable experience for users during multi-threaded activities.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -6,6 +6,7 @@ import json
 import os
 import re
 import sys
+import threading
 from collections.abc import Callable, Generator
 from io import StringIO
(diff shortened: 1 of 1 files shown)


sarahbx:
moved location of `import threading` to match current import format

myakove (approved):

myakove:
@sarahbx 
please run `tox` locally, this will run functional tests.

PR #2372, opened 2025-04-01
Revert "Update wait_for_status_success() call to look at both type and status for status.conditions"

myakove:
Reverts RedHatQE/openshift-python-wrapper#2289

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the network configuration status check for faster and more reliable detection of successful configurations, enhancing overall responsiveness.
  - Improved logging for better traceability of status updates.


myakove:
> Does not look like a clean revert as it is removing other people's changes as well. Specially after almost 3 months of the merge, may be fixing what does not work would be more simple?

Thanks, this is only for testing.

PR #2391, opened 2025-04-29
feat: allow resource.get to accept Namespace instance instead

sheltoncyril:
##### Short description:
This feature allows Namespace objects to be passed into the get function of NamespacedResources. Pod.get is an example which can directly accept a namespace avoiding confusion of where Namespace exists as a string.

##### More details:
It's annoying to juggle between namespaces as strings and namespace objects. If the Namespace is passed into a function such as Pod.get it results in the field resolving to __repr__ which is the default behavior. 

I think this fixes two things but the implementation is mostly a suggestion. 
- When str, format or print is called on a Namespace instance, it resolves to a human-readable string.
- When a namespace is passed into a getter function, the name is used rather than the repr.

I'm quite new to the team so I hope you can give me more detailed guidance and steps to improve.
I really appreciate detailed, honest feedback!

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -1368,7 +1368,7 @@ class NamespacedResource(Resource):
                         yield cls(
                             client=dyn_client,
                             name=resource_field.metadata.name,
-                            namespace=resource_field.metadata.namespace,
+                            namespace=str(resource_field.metadata.namespace),
                         )
             except TypeError:
                 if raw:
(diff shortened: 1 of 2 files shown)


sheltoncyril:
pre-commit.ci autofix

ocp_resources/resource.py
```diff
@@ -1368,7 +1368,7 @@ def _get() -> Generator["NamespacedResource|ResourceInstance", None, None]:
                         yield cls(
                             client=dyn_client,
                             name=resource_field.metadata.name,
-                            namespace=resource_field.metadata.namespace,
+                            namespace=str(resource_field.metadata.namespace),
```
dbasunag, line 1371: This issue seems like an user error situation. namespace here is namespace name. This is true for all wrapper objects, if I recall correctly. Just like an oc command does not work with kinds as params, wrapper expects namespace strings when you are providing a namespace.

PR #2467, opened 2025-07-23
docs: update install instructions to use uv in README

SamAlber:
This PR updates the README to reflect the correct installation using uv with `pyproject.toml`
* `setup.py` doesn't exist.

## Summary by CodeRabbit

* **Documentation**
  * Updated installation instructions to unify commands under the `uv` tool.
  * Added guidance for local installation and usage of the wrapper in other projects.
  * Revised PyPI installation command for improved clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,19 @@ From source:
 ```bash
 git clone https://github.com/RedHatQE/openshift-python-wrapper.git
 cd openshift-python-wrapper
-python setup.py install --user
+uv sync
 ```
 
-From pypi:
+To use the wrapper in another project:
 
 ```bash
-pip install openshift-python-wrapper --user
+uv pip install /path/to/openshift-python-wrapper
+```
+
+From Pypi:
+
+```bash
+uv pip install openshift-python-wrapper
 ```
 
 ## Fake Kubernetes Client


README.md
```diff
@@ -32,7 +32,7 @@ From source:
 ```bash
 git clone https://github.com/RedHatQE/openshift-python-wrapper.git
 cd openshift-python-wrapper
-python setup.py install --user
+pip install . # add -e if you're making local changes
```
myakove, line 35: Thanks for this PR.
Please instruct to use `uv`

myakove (changes requested):

README.md
```diff
@@ -32,7 +32,7 @@ From source:
 ```bash
 git clone https://github.com/RedHatQE/openshift-python-wrapper.git
 cd openshift-python-wrapper
-python setup.py install --user
+uv pip install . # Add -e for editable/development mode if making local changes
```
myakove, line 35: I think there is no point install it on the same folder under same venv of uv (uv sync will have the code anyway)
The only option I think is valid is if someone want to develope and test changes in the wrapper with another project, in this case it should be something like:

git clone https://github.com/RedHatQE/openshift-python-wrapper.git
cd openshift-python-wrapper
pwd
and in the other project do uv/pip/.... install ${pwd}.

WDYT?

PR #2499, opened 2025-08-13
refactor(wait_for_status): Use self.instance to get status

myakove:
## Summary by CodeRabbit

- New Features
  - None

- Improvements
  - More reliable and efficient resource status polling during waits.
  - Reduced logging noise by only reporting status changes.
  - Early exit when encountering terminal failure statuses.
  - Clearer timeout messages that include the last-known status.

- Bug Fixes
  - Fixes issues from parsing list-style responses for status.
  - Prevents redundant logs during long-running waits.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -970,27 +970,23 @@ class Resource(ResourceConstants):
                 **PROTOCOL_ERROR_EXCEPTION_DICT,
                 **DEFAULT_CLUSTER_RETRY_EXCEPTIONS,
             },
-            func=self.api.get,
-            field_selector=f"metadata.name=={self.name}",
-            namespace=self.namespace,
+            func=lambda: self.instance.status.phase,
         )
         current_status = None
         last_logged_status = None
         try:
             for sample in samples:
-                if sample.items:
-                    sample_status = sample.items[0].status
-                    if sample_status:
-                        current_status = sample_status.phase
-                        if current_status != last_logged_status:
-                            last_logged_status = current_status
-                            self.logger.info(f"Status of {self.kind} {self.name} is {current_status}")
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -970,27 +970,23 @@ def wait_for_status(
                 **PROTOCOL_ERROR_EXCEPTION_DICT,
                 **DEFAULT_CLUSTER_RETRY_EXCEPTIONS,
             },
-            func=self.api.get,
-            field_selector=f"metadata.name=={self.name}",
-            namespace=self.namespace,
+            func=lambda: self.instance.status.phase,
```
rnetser, line 973: can we fail here in cases where there's no `status.phase` like "None has no attirbute..."?

vsibirsk: how about 
```
func=lambda: self.exists,
```
and 
```
if sample:
    current_status = sample.status.phase
```

vsibirsk (approved):

ocp_resources/resource.py
myakove: Switch to use `self.exists`

ocp_resources/resource.py
```diff
@@ -971,27 +971,25 @@ def wait_for_status(
                 **PROTOCOL_ERROR_EXCEPTION_DICT,
                 **DEFAULT_CLUSTER_RETRY_EXCEPTIONS,
             },
-            func=self.api.get,
-            field_selector=f"metadata.name=={self.name}",
-            namespace=self.namespace,
+            func=lambda: self.exists,
```
dbasunag, line 974: this would work and you can also use self.instance.status here and compare it in line 980.

PR #2513, opened 2025-08-21
feat: Retry MR creation on webhook error

dbasunag:
## Summary by CodeRabbit

- New Features
  - Added a create method to the Model Registry, enabling users to create resources with an optional wait-for-completion option.

- Bug Fixes
  - Improved reliability of resource creation by automatically retrying transient cluster errors and applying sensible retry/timeouts to reduce failures during intermittent connection issues.

### DIFF
--- a/ocp_resources/model_registry_modelregistry_opendatahub_io.py
+++ b/ocp_resources/model_registry_modelregistry_opendatahub_io.py
@@ -1,7 +1,9 @@
 # Generated using https://github.com/RedHatQE/openshift-python-wrapper/blob/main/scripts/resource/README.md
-
-
 from typing import Any
+
+from kubernetes.dynamic import ResourceInstance
+from kubernetes.dynamic.exceptions import InternalServerError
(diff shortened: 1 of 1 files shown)


ocp_resources/model_registry_modelregistry_opendatahub_io.py
```diff
@@ -92,3 +94,23 @@ def to_dict(self) -> None:
                 _spec["postgres"] = self.postgres
 
     # End of generated code
+
+    def create(self, wait: bool = False) -> ResourceInstance | None:
```
myakove, line 98: I don't think this is the place to do it.
do the retry when you call create()

myakove: `["connection refused", "no endpoints available for service"]` Seems like a valid error, I don't know how the flow should work, but maybe you need to make sure the endpoint is available before this resource?

PR #2519, opened 2025-09-03
Add openTelemetry Collector resource

kpunwatk:
Adds openTelemetry resource to the wrapper

## Summary by CodeRabbit

- New Features
  - Added support for the OpenTelemetry Collector custom resource, including fields for config, management state, mode, replicas, and resources.
  - Enforced validation when generating resource definitions, raising errors if required fields are missing.
  - Exposed a new API group for OpenTelemetry to align with the custom resource.

- Documentation
  - Updated public API surface to include the new resource and API group.


kpunwatk:
Sorry would open a new PR

PR #2540, opened 2025-10-17
fix: incluster config handling

dtrifiro:
The current implementation is broken if run in a cluster.

It should be possible to load an in-cluster configuration via `kubernetes.config.load_incluster_config()`, which populates
host/ports/certs from environment variables and secrets from `/var/run/secrets/kubernetes.io`.

There's a few issues in the current implementation:

- Running `get_client()` falls through `new_client_from_config` which raises `ConfigException: Invalid kube-config file. No configuration found.`,
- The final return is broken, since the `DynamicClient` is instantiated with `client=kubernetes.config.incluster_config.load_incluster_config`, which is always None
- Trying to bypass the above problems by creating a custom `client_config` and passing it to `get_client` also fails: it falls through to `new_client_from_config` and raises `ConfigException`, because `load_incluster_config` sets `client_config.api_key="bearer <token>"` which is ignored.

This PR fixes the issues by:
- only attempting to load the configuration file if passed explicitly via arg, explicitly set via `KUBECONFIG` env var or if `~/.kube.config` exists.
- loading the incluster config if in a k8s environment (`KUBERNETES_SERVICE_PORT` and `KUBERNETES_SERVICE_HOST` are set)
- fall through passing the `client_configuration` as is, fixing the last issue. 

## Summary by CodeRabbit

* **New Features**
  * Automatic environment-driven client selection: prefers KUBECONFIG, falls back to default kubeconfig, and detects in-cluster configuration.

* **Bug Fixes**
  * More reliable initialization flow with clearer selection logic and explicit error when no valid configuration is found; in-cluster handling no longer relies on brittle fallback heuristics.

* **Chores**
  * Improved informational logging about which configuration source is used; no public API signature changes.

### DIFF
--- a/ocp_resources/resource.py
+++ b/ocp_resources/resource.py
@@ -35,7 +35,6 @@ from timeout_sampler import (
     TimeoutSampler,
     TimeoutWatch,
 )
-from urllib3.exceptions import MaxRetryError
 
 from fake_kubernetes_client.dynamic_client import FakeDynamicClient
 from ocp_resources.event import Event
@@ -204,16 +203,24 @@ def get_client(
     token: str | None = None,
     fake: bool = False,
 ) -> DynamicClient | FakeDynamicClient:
-    """
-    Get a kubernetes client.
+    """Get a kubernetes client.
+
+    Valid combinations for arguments include:
+        - config file (kubeconfig path)
+        - config_dict
+        - username, password and host (uses basic auth)
+        - host and token (bearer token)
+        - client_configuration
+
+    If none of the above arguments are provided:
+    - if the `KUBECONFIG` env var is set: it will be used to configure the client.
+    - if `KUBECONFIG` is not set and `~/.kube/config` exists, it will be used to configure the client.
+    - if `KUBECONFIG` is not set and `~/.kube/config` does not exist, in-cluster configuration will
+      be attempted when running inside a Kubernetes cluster.
 
 
-    This function is a replica of `ocp_utilities.infra.get_client` which cannot be imported as ocp_utilities imports
-    from ocp_resources.
+    If no configuration source is found, kubernetes.config.ConfigException will be raised.
 
-    Pass either config_file or config_dict.
-    If none of them are passed, client will be created from default OS kubeconfig
-    (environment variable or .kube folder).
 
     Args:
         config_file (str): path to a kubeconfig file.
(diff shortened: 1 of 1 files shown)


ocp_resources/resource.py
```diff
@@ -265,33 +264,31 @@ def get_client(
...
-        # is populated during import which comes before setting the variable in code.
-        config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
-
-        _client = kubernetes.config.new_client_from_config(
-            config_file=config_file,
-            context=context,
-            client_configuration=client_configuration,
-            persist_config=persist_config,
-        )
+        try:
```
myakove: the code should reflect my comment in line 214, only if none of them passed we should do client from incluster config

ocp_resources/resource.py
```diff
@@ -264,34 +267,38 @@ def get_client(
...
-        # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will be None since
-        # is populated during import which comes before setting the variable in code.
-        config_file = config_file or os.environ.get("KUBECONFIG", "~/.kube/config")
+        # If `KUBECONFIG` environment variable is set via code, the `KUBE_CONFIG_DEFAULT_LOCATION` will not be updated
+        # to reflect the new value since it's set to at import-time.
+
+        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
+        if not config_file:
+            config_file = os.environ.get("KUBECONFIG", "~/.kube/config")
+        LOGGER.info("Trying to get client via new_client_from_config config_file=%s", config_file)
```
myakove: Please use f-string and not `%s`

ocp_resources/resource.py
```diff
@@ -264,34 +267,38 @@ def get_client(
...
 
         _client = kubernetes.config.new_client_from_config(
             config_file=config_file,
             context=context,
             client_configuration=client_configuration,
             persist_config=persist_config,
         )
+    elif all(
```
myakove: Any reason why not query `os.environ` with the needed keys?

ocp_resources/resource.py
dtrifiro: This is the correct way to pass argument to logging statements: formatting of message arguments is deferred until it cannot be avoided.

https://docs.python.org/3/howto/logging.html#optimization
https://docs.python.org/3/howto/logging.html#logging-variable-data

ocp_resources/resource.py
```diff
@@ -264,34 +267,38 @@ def get_client(
             persist_config=persist_config,
             temp_file_path=temp_file_path,
         )
-    else:
-        # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/__init__.py
-        LOGGER.info("Trying to get client via new_client_from_config")
-
+    elif config_file or ("KUBECONFIG" in os.environ) or os.path.exists(os.path.expanduser("~/.kube/config")):
```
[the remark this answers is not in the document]
dtrifiro: `TEST_KUBECONFIG` as an environment variable? In that case `"KUBECONFIG" in os.environ` will not be true, `os.environ` is a dict and will not have the `"KUBECONFIG"` key.

The idea here is to go through this path only if:
- `config_file` is passed by the user explicitly
- user explicitly set a `KUBECONFIG` environment variable (so the key will be in `os.environ`)
- a kubeconfig is available at the default location

myakove (changes requested):

ocp_resources/resource.py
```diff
@@ -265,35 +269,46 @@ def get_client(
...
-    kubernetes.client.Configuration.set_default(default=client_configuration)
-
-    try:
-        return kubernetes.dynamic.DynamicClient(client=_client)
-    except MaxRetryError:
         # Ref: https://github.com/kubernetes-client/python/blob/v26.1.0/kubernetes/base/config/incluster_config.py
-        LOGGER.info("Trying to get client via incluster_config")
-        return kubernetes.dynamic.DynamicClient(
-            client=kubernetes.config.incluster_config.load_incluster_config(
-                client_configuration=client_configuration, try_refresh_token=try_refresh_token
```
myakove: Please send `try_refresh_token=try_refresh_token` in your code

dtrifiro:
@myakove done.

PR #2599, opened 2025-12-05
fix: Use the right name for PostHook

krcmarik:
There is no such type of hook as AfterHook used in the Plan resource. The correct type is PostHook.
hook_type='AfterHook' is not a valid Forklift API value and never was (only PreHook and PostHook are valid). Any code using after_hook_name would fail with HookStepNotValid error.

## Summary by CodeRabbit

* **New Features**
  * Plans now support pre-hook and post-hook execution points.

* **Breaking Changes**
  * The previous "after-hook" identifier has been renamed to "post-hook". Update configurations and automations to use post-hook names to maintain behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
           ]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.7
+    rev: v0.14.8
     hooks:
       - id: ruff
       - id: ruff-format
--- a/uv.lock
+++ b/uv.lock
@@ -661,7 +661,7 @@ wheels = [
 
 [[package]]
 name = "fastmcp"
-version = "2.13.2"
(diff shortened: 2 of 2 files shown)


ocp_resources/plan.py
```diff
@@ -58,8 +62,8 @@ def __init__(
         warm_migration: bool = False,
         pre_hook_name: str | None = None,
         pre_hook_namespace: str | None = None,
-        after_hook_name: str | None = None,
-        after_hook_namespace: str | None = None,
+        post_hook_name: str | None = None,
```
rnetser: this is a breaking change
you can add the new args but not drop the old ones (and use where needed) and add deprecation warning on the old ones.

rnetser (changes requested):

PR #2649, opened 2026-02-13
feat: Add future warning for api_name in DataVolume constructor, it will default to storage

ema-aka-young:
##### Short description:
Add a warning to inform consumers that the default API name will soon change from "pvc" to "storage".

##### More details:
Implementing step 1 of https://issues.redhat.com/browse/CNV-79539

We need this to achieve a smooth transition to "storage" as the default API name when creating a DataVolume.

## Summary by CodeRabbit

* **Chores**
  * DataVolume now emits a deprecation warning when the `api_name` parameter is not explicitly specified. The default value will change from "pvc" to "storage" in a future release. Explicitly set `api_name` to avoid warnings and ensure compatibility with future versions.


ema-aka-young:
/verified
through `ipython` and `--with-editable ../openshift-python-wrapper`

PR #2759, opened 2026-07-09
feat: add aap_job_template_id support to Hook class

AmenB:
## Summary

Add `aap_job_template_id` parameter to the `Hook` class, enabling MTV AAP hook integration — Hook CRs that trigger AWX/AAP job templates during migration instead of running playbooks in containers.

## What changed

- **New parameter**: `aap_job_template_id=None` on `Hook.__init__()`.
- **Branching logic in `to_dict()`**:
  - When `aap_job_template_id` is set → produces `spec.aap.jobTemplateId`.
  - Otherwise → produces `spec.image` + `spec.playbook` (existing behavior, unchanged).

No regression — existing callers that don't pass `aap_job_template_id` hit the original code path.

## Why

The `mtv-api-tests` project (branch `feat/aap-hook-integration`) has a new AAP hook integration test that creates a Hook CR via `Hook(aap_job_template_id=9)`. Without this wrapper change, it fails with:

```
TypeError: Resource.__init__() got an unexpected keyword argument 'aap_job_template_id'
```

Confirmed by running the test against the current PyPI release (11.0.134).

## Related Jira tickets

- **MTV-6031**: [Automation] AAP Hook Integration test
- **MTV-3663**: [TP] MTV hook for AAP integration (feature epic)
- **MTV-6063**: AAP hook migration hangs on TLS error (bug found during testing)

**Polarion**: MTV-781

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Hook resources via an AAP template ID (`aap`), plus `deadline`, `image`, `playbook`, and `service_account`.
* **Behavior Changes**
  * Hook output no longer forcibly injects `spec.image` / `spec.playbook`; serialization now follows the provided input parameters.
  * `image` defaulting is conditional: it’s omitted when `aap` is set, otherwise defaults apply.
* **Deprecations**
  * A deprecation warning is now always shown when using the legacy Hook interface.

### DIFF
--- a/ocp_resources/hook.py
+++ b/ocp_resources/hook.py
@@ -1,49 +1,42 @@
-from ocp_resources.resource import NamespacedResource
-from ocp_resources.utils.constants import TIMEOUT_4MINUTES
+import warnings
+from typing import Any
+
+from ocp_resources._hook_generated import Hook as _GeneratedHook
+
+_UNSET = object()
+_DEFAULT_IMAGE = "quay.io/konveyor/hook-runner:latest"
 
 
-class Hook(NamespacedResource):
+class Hook(_GeneratedHook):
     """
-    Migration Tool for Virtualization (MTV) Plan's Hook Resource.
-    """
+    Deprecated shim for backward compatibility.
 
-    api_group = NamespacedResource.ApiGroup.FORKLIFT_KONVEYOR_IO
+    Preserves the legacy default-image behavior. New code should use
+    ``from ocp_resources._hook_generated import Hook`` directly.
+    """
 
     def __init__(
         self,
-        name=None,
-        namespace=None,
-        image="quay.io/konveyor/hook-runner:latest",
-        playbook=None,
-        client=None,
-        teardown=True,
-        yaml_file=None,
-        delete_timeout=TIMEOUT_4MINUTES,
-        **kwargs,
-    ):
-        """
-        Args:
-            image (str): Path to an ansible image
-            playbook (str): Ansible playbook to be performed
-        """
+        aap: dict[str, Any] | None = None,
+        deadline: int | None = None,
+        image: Any = _UNSET,
+        playbook: str | None = None,
+        service_account: str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        warnings.warn(
+            "Hook legacy defaults are deprecated and will be removed. "
+            "Pass image= explicitly for local hooks; use aap={...} for AAP hooks.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        if image is _UNSET:
+            image = None if aap is not None else _DEFAULT_IMAGE
         super().__init__(
-            name=name,
-            namespace=namespace,
-            client=client,
-            teardown=teardown,
-            yaml_file=yaml_file,
-            delete_timeout=delete_timeout,
+            aap=aap,
+            deadline=deadline,
+            image=image,
+            playbook=playbook,
+            service_account=service_account,
             **kwargs,
         )
-        self.image = image
-        self.playbook = playbook
-
-    def to_dict(self) -> None:
-        super().to_dict()
-        if not self.kind_dict and not self.yaml_file:
-            self.res.update({
-                "spec": {
-                    "image": self.image,
-                    "playbook": self.playbook,
-                },
-            })
--- a/tests/scripts/validate_resources.py
+++ b/tests/scripts/validate_resources.py
@@ -106,7 +106,7 @@ def validate_resource(
 
 
 def resource_file() -> Generator[str, None, None]:
-    ocp_resources_exclude_files = ["resource.py", "utils.py", "__init__.py"]
+    ocp_resources_exclude_files = ["resource.py", "utils.py", "__init__.py", "hook.py"]
     for root, _, files in os.walk("ocp_resources"):
         for _file in files:
             if _file in ocp_resources_exclude_files or not _file.endswith(".py"):


ocp_resources/hook.py
```diff
@@ -37,13 +39,23 @@ def __init__(
...
     def to_dict(self) -> None:
         super().to_dict()
         if not self.kind_dict and not self.yaml_file:
-            self.res.update({
-                "spec": {
-                    "image": self.image,
-                    "playbook": self.playbook,
-                },
-            })
+            if self.aap_job_template_id is not None:
```
myakove: Why do we have here if;else?

myakove: and if you already modify the code here, switch to use the standatd approch for doing this
self.res["spec"] = {}
 _spec = self.res["spec"]
if self.aap_job_template_id is not None:
    _spec[jobTemplateId] = self.aap_job_template_id 

for both please

AmenB: The Hook CRD supports two mutually exclusive modes — playbook-based (spec.image + spec.playbook) and 
AAP-based (spec.aap.jobTemplateId). They can't coexist in the same CR, so the if/else picks which spec
shape to emit

myakove: we do not block anything in code level, if it can sent to ocp api we send it, the ocp api will give the error if there is one.

AmenB: The if/else is needed here because `image` has a default value (`"quay.io/konveyor/hook-runner:latest"`), so it's always truthy. Without the if/else, AAP hooks end up with both `spec.aap.jobTemplateId` and `spec.image` in the CR — Forklift sees both and the Plan never reaches Ready.

We tested locally without the if/else and confirmed it fails. The two hook types have different spec shapes — it's not validation, just picking which fields to emit based on what the caller asked for.

I did switch to the standard `_spec` pattern as you suggested.

myakove: @rnetser thinkg of removing the defult image, WDYT?

myakove: @AmenB for now keep the if else and add a detraction warn about the `image` default will be gone in next release

AmenB: @myakove Added the deprecation warning. Note that it'll fire for every playbook hook that doesn't pass image explicitly — could be noisy. Want a different approach?

ocp_resources/hook.py
```diff
@@ -35,15 +39,23 @@ def __init__(
...
         )
+        if image == "quay.io/konveyor/hook-runner:latest":
+            warnings.warn(
+                "The default value for 'image' will be removed in the next release. Pass 'image' explicitly.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
         self.image = image
         self.playbook = playbook
+        self.aap_job_template_id = aap_job_template_id
```
myakove: Use a sentinel to distinguish "caller didn't pass `image`" from "caller explicitly passed the default value." String comparison can't tell the two apart — someone calling `Hook(image="quay.io/konveyor/hook-runner:latest")` explicitly still gets warned.

```suggestion


_UNSET = object()


        if image is _UNSET:
           
                warnings.warn(
                    "The default value for 'image' will be removed in the next release. Pass 'image' explicitly.",
                    DeprecationWarning,
                    stacklevel=2,
                )
            image = "quay.io/konveyor/hook-runner:latest" if aap_job_template_id is None else None


```

AmenB: Done, used the sentinel pattern and switched to the _spec approach.

redhat-qe-bot2:
@AmenB is now allowed to run commands

ocp_resources/hook.py
```diff
@@ -35,15 +41,24 @@ def __init__(
...
-                "spec": {
-                    "image": self.image,
-                    "playbook": self.playbook,
-                },
-            })
+            self.res["spec"] = {}
+            _spec = self.res["spec"]
+            if self.aap_job_template_id is not None:
+                _spec["aap"] = {"jobTemplateId": self.aap_job_template_id}
+            else:
```
myakove: Please add a TDOD: to remove the else once `image` default is deprecated.

AmenB: Done.

myakove (dismissed):

ocp_resources/hook.py
```diff
@@ -35,15 +41,25 @@ def __init__(
...
-            self.res.update({
-                "spec": {
-                    "image": self.image,
-                    "playbook": self.playbook,
-                },
-            })
+            self.res["spec"] = {}
+            _spec = self.res["spec"]
+            if self.aap_job_template_id is not None:
+                _spec["aap"] = {"jobTemplateId": self.aap_job_template_id}
```
rnetser: @myakove I think that we should get here `app` (as we do now with new generated resources). wdyt?

ocp_resources/hook.py
```diff
@@ -35,15 +41,25 @@ def __init__(
...
         super().to_dict()
         if not self.kind_dict and not self.yaml_file:
-            self.res.update({
-                "spec": {
-                    "image": self.image,
-                    "playbook": self.playbook,
-                },
-            })
+            self.res["spec"] = {}
+            _spec = self.res["spec"]
```
rnetser: can `app` and `playbook` \ `image` co-exist? iiuc, they cannot. while we do not validate that (the api should); we have opinionated ordering here - i.e if a user passes both, we give `app` presedence. @myakove wdyt?

rnetser: i mean in other places, for new resources, we do not have the if-else option but let the api raise if a user passes mutually exlusive values, should we do it here as well?

myakove: No, we send what the caller passing us.

myakove:
@AmenB 

We're landing the schema-aligned generated Hook in a separate PR: https://github.com/RedHatQE/openshift-python-wrapper/pull/2765 (ocp_resources/_hook_generated.py  
 — fields: aap, deadline, image, playbook, service_account). That PR does not change hook.py or add deprecation.                                                     
                                                                                                                                                                     
 Once #2765 (https://github.com/RedHatQE/openshift-python-wrapper/pull/2765) merges, please retarget this PR to:                                                     
                                                                                                                                                                     
 1. Keep from ocp_resources.hook import Hook working by turning hook.py into a thin deprecated shim that subclasses / delegates to _hook_generated.Hook.             
 2. Do not add aap_job_template_id. Use the generated aap: dict instead, e.g. Hook(..., aap={"jobTemplateId": 9}).                                                   
 3. Keep the legacy default-image behavior in the shim only (same idea as _UNSET here): if image is omitted and aap is set → image=None; if image is omitted and aap 
     is not set → apply the current default image + DeprecationWarning.                                                                                              
 4. Remove the if/else in to_dict. The generated class already writes only non-None fields (aap / image / playbook / …).                                             
                                                                                                                                                                     
 Example shim shape:                                                                                                                                                 
                                                                                                                                                                     
 ```python                                                                                                                                                           
   from ocp_resources._hook_generated import Hook as _GeneratedHook                                                                                                  
                                                                                                                                                                     
   _UNSET = object()                                                                                                                                                 
   _DEFAULT_IMAGE = "quay.io/konveyor/hook-runner:latest"                                                                                                            
                                                                                                                                                                     
   class Hook(_GeneratedHook):                                                                                                                                       
       def __init__(self, aap=None, deadline=None, image=_UNSET, playbook=None, service_account=None, **kwargs):                                                     
           warnings.warn(                                                                                                                                            
               "Hook legacy defaults are deprecated and will be removed. "                                                                                           
               "Pass image= explicitly for local hooks; use aap={...} for AAP hooks.",                                                                               
               DeprecationWarning,                                                                                                                                   
               stacklevel=2,                                                                                                                                         
           )                                                                                                                                                         
           if image is _UNSET:                                                                                                                                       
               image = None if aap is not None else _DEFAULT_IMAGE                                                                                                   
           super().__init__(                                                                                                                                         
               aap=aap,                                                                                                                                              
               deadline=deadline,                                                                                                                                    
               image=image,                                                                                                                                          
               playbook=playbook,                                                                                                                                    
               service_account=service_account,                                                                                                                      
               **kwargs,                                                                                                                                             
           )                                                                                                                                                         
 ```                                                                                                                                                                 
                                                                                                                                                                     
 MTV call site change: Hook(aap_job_template_id=9) → Hook(aap={"jobTemplateId": 9}).

redhat-qe-bot1:
AmenB is not part of the approver, only approvers can mark pull request with hold

ocp_resources/hook.py
```diff
@@ -1,49 +1,42 @@
...
+        **kwargs: Any,
+    ) -> None:
+        warnings.warn(
+            "Hook legacy defaults are deprecated and will be removed. "
+            "Pass image= explicitly for local hooks; use aap={...} for AAP hooks.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        if image is _UNSET:
+            image = None if aap is not None else _DEFAULT_IMAGE
```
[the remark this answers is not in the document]
AmenB: Keeping the warning outside the if block intentionally — maintainer wants all shim callers to know this class is deprecated, even if they already pass image= explicitly.

rh-bot-1:
**Clean rebase detected** — no code changes compared to previous head (`15bbb54`).
The following labels were preserved: `commented-coderabbitai[bot]`, `commented-AmenB`.

PR #2801, opened 2026-08-31
feat: Add TargetReady to resource status constants

Anatw:
##### Short description:
The migration TargetReady phase had no shared constant. Add it alongside the existing status phase constants so consumers can reference it directly.

This status is needed to track a VM live migration phase.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing the `TargetReady` resource status.


Anatw:
/verified

Constant addition - nothing to run.

